PR #9, opened 2023-08-01
Planned 0.1.0 breaking changes

paf31:
- [x] Rename Table to Collection
- [x] Rename Command to Procedure
- [x] Update `RowSet` type to avoid additional wrapping (`{ "value": ... }`)
- [x] Combine `PathElement` and `PathElementWithPredicate`
- [x] Reference implementation for procedures
- [x] Preserve ordering of fields in all selection-set-like response objects
- [x] Change `ExplainResponse` to consist of unconstrained key value pairs


ndc-client/tests/json_schema/schema_response.jsonschema
```diff
@@ -0,0 +1,465 @@
...
+        },
+        "foreign_keys": {
+          "description": "Any foreign key constraints enforced on this collection",
+          "type": "object",
+          "additionalProperties": {
+            "$ref": "#/definitions/ForeignKeyConstraint"
+          }
+        },
+        "insertable_columns": {
+          "description": "The set of names of insercollection columns, or null if inserts are not supported",
```
SamirTalwar, line 127: I think your find/replace went haywire here.

paf31: Argh I tried so hard to avoid this exact replacement 😠

PR #18, opened 2023-08-14
Set up Nix to provide a development shell and build the tools.

SamirTalwar:
I'm using this for the development shell, but I set up the build targets for the applications anyway.


paf31:
I don't personally use Nix or know the language, so I can't review the code itself, and I am not going to maintain anything related to Nix that we add to this repository. At a minimum, I'd want to try to exercise this code in CI somehow. Can that be done?

SamirTalwar:
Yeah, we can definitely do that. I'll take a crack at it tomorrow.

paf31:
Is there a reason this needs to be in `main` before it's ready in CI? As I say, I don't mind adding it, but I want CI so that I can be sure I won't break it myself. 

I don't use Nix but I have to assume there's a way to provide a script for something without committing it to that thing's main repo?

SamirTalwar:
@paf31, I got CI working on GitHub Actions.

paf31 (approved):

PR #19, opened 2023-08-14
Fix the tests; the JSON key sort order has changed.

SamirTalwar:


paf31:
Thanks! Not sure how this went wrong, possibly we need to change the merge policy to rebase-before-merge.

PR #20, opened 2023-08-14
Fix all warnings flagged by Clippy.

SamirTalwar:
Seemed sensible, that's all.

### DIFF
--- a/ndc-reference/bin/reference/main.rs
+++ b/ndc-reference/bin/reference/main.rs
@@ -11,7 +11,7 @@ use axum::{
     routing::{get, post},
     Json, Router,
 };
-use csv;
+
 use indexmap::IndexMap;
 use ndc_client::models;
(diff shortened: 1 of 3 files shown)


ndc-reference/bin/reference/main.rs
```diff
@@ -1014,17 +982,17 @@ fn eval_path_element(
...
                     state,
-                    &predicate,
+                    predicate,
                     tgt_row,
                     tgt_row,
-                )? {
-                    matching_rows.push(tgt_row.clone());
-                }
+                )?
+            {
```
paf31, line 994: Is this just formatting? Does clippy differ from whatever rust-analyzer does to format (I'm assuming `cargo fmt`)?

SamirTalwar: The change is from a nested `if` to using `&&`, which means one fewer nested brace pair.

SamirTalwar:
I think that'd be a good idea.

PR #21, opened 2023-08-14
Turn ndc-test into a library with an optional binary.

SamirTalwar:
This means we can run it in two ways: either it can be run as a binary, or pulled into an integration test.

To do this, I pulled ndc-test into its own top-level crate.


paf31 (approved):

paf31:
@sordina We can now make a plan to add `ndc-test` support to `ndc-sdk` in some form.

paf31:
Thanks @SamirTalwar!

PR #25, opened 2023-08-27
Updated test runner

paf31:
- Adds some property-based tests which infer generators using the data returned by the query API.
- Removes the last vestige of column-level arguments (missed in the previous PR)
- Improves error reporting to use `Result` instead of `assert`/`panic`.

### DIFF
--- a/ndc-client/src/apis/configuration.rs
+++ b/ndc-client/src/apis/configuration.rs
@@ -7,7 +7,6 @@ pub struct Configuration {
     pub oauth_access_token: Option<String>,
     pub bearer_access_token: Option<String>,
     pub api_key: Option<ApiKey>,
-    pub seed: Option<String>,
 }
 
 pub type BasicAuth = (String, Option<String>);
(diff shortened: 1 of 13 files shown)


ndc-client/src/apis/configuration.rs
```diff
@@ -7,6 +7,7 @@ pub struct Configuration {
     pub oauth_access_token: Option<String>,
     pub bearer_access_token: Option<String>,
     pub api_key: Option<ApiKey>,
+    pub seed: Option<String>,
```
paf31, line 10: I didn't realize this had changed this file, I will fix this.

soupi:
Hey Phil, nice work. I ran this on the postgres-ndc and got a few errors, for example:

```
[1] Predicates
  in Employee
  in Simple queries
Details: error communicating with the connector: error in response: status code 500 Internal Server Error
```

However, this error does not contain enough information for me to act on. What would help me is a way to reproduce the failing test, for example one of the following:

1. Show me the `QueryRequest`
2. Let me rerun the failing test only, so I can look at the `QueryRequest` in the connector's logs
3. Provide a `curl` command to reproduce the failing test

paf31:
Thanks @soupi. I have some follow up work which will add the ability to save all generated request/response pairs as snapshots, so I think that will partially solve the issue you're seeing, but I do think we can improve the error messages. I would like to save that for a follow up PR though, if that works for you.

codedmart (approved):

PR #26, opened 2023-08-31
add function and procedure implementations

purugupta99:
This PR adds implementation for some functions and procedure to be used for testing of commands in `v3-engine`. We add the following:

#### Functions
- `latest_article_id`: this function will be called without any argument and returns a scalar type `Int`
- `latest_article`: this function will be called without any arguments and return a custom type `article`
- `get_all_articles`: this function will be called without any arguments and return a custom type `article` wrapped in a container `array`
- `get_article_by_id`: this function needs to be called with an argument `id` and return a custom type `article`

#### Procedures
- `update_article_title_by_id`: this procedure takes in two arguments `id` and `title`, updates the article if the given `id` exists in the source with the provided `title` and returns back the updated article and number of rows affected.

These have been added to verify the working of commands and write tests for various scenarios of argument & output types supported by commands.

Another change made as part of this PR is updating `execute_query` function to handle the case when the response is generated for a function as it has an extra `__value` key in the response root.


paf31 (commented):
Can you please say a bit about why these are necessary?

Also, we should add tests for any new functions, and update the generated documentation (`specification/book`), but let's hold off on that until we clarify which functions we need here, because this might also bloat the generated documentation.

purugupta99:
> Can you please say a bit about why these are necessary?
> 
> Also, we should add tests for any new functions, and update the generated documentation (`specification/book`), but let's hold off on that until we clarify which functions we need here, because this might also bloat the generated documentation.

Updated the PR description with the same, let me know if we need more info. Once the functions are confirmed I will add the tests and docs.

paf31:
Ok, I see. Could you please add test cases for the new functions in that case. We don't need to include all of the new examples in the tutorial docs, but we should make sure the text is still valid, and that we regenerate everything, and that it renders okay, please.

ndc-reference/bin/reference/main.rs
```diff
@@ -525,6 +608,140 @@ fn get_collection_by_name(
...
+                        Ok::<_, StatusLine>(rows)
+                    })
+                    .transpose()?;
+
+                let row_set = models::RowSet {
+                    aggregates: None,
+                    rows,
+                };
+
+                let latest_article_value = serde_json::to_value(row_set)
```
paf31, line 651: Why does this return a `RowSet`? The return type is `article`.

ndc-reference/bin/reference/main.rs
```diff
@@ -525,6 +608,140 @@ fn get_collection_by_name(
...
+                    }
+                    Ok::<_, StatusLine>(rows)
+                })
+                .transpose()?;
+
+            let row_set = models::RowSet {
+                aggregates: None,
+                rows,
+            };
+            let articles_value = serde_json::to_value(row_set)
```
paf31, line 689: Same issue

ndc-reference/bin/reference/main.rs
```diff
@@ -396,8 +428,59 @@ async fn get_schema() -> Json<models::SchemaResponse> {
...
+            underlying_type: Box::new(models::Type::Named {
+                name: "article".into(),
+            }),
+        },
+    };
+    // ANCHOR_END: schema_function_get_article_by_id
+
     // ANCHOR: schema_functions
-    let functions: Vec<models::FunctionInfo> = vec![latest_article_id_function];
+    let functions: Vec<models::FunctionInfo> = vec![
```
paf31, line 478: This bit gets included in the spec, and the same for `let procedures = ...`. We ought to include a change to the text to explain that these functions exist to illustrate various types of functions, and that we won't go into detail about each one in the text.

paf31 (commented):
Now that I see all of the changes here together, I'm thinking it's a big change and a lot of new code which debatably is not really needed in the executable specification. If we need all of these functions for testing purposes, then it might be better to maintain different connectors for specific testing purposes, because the purpose of the reference implementation is not only for testing.

At the very least I think we should think about splitting these functions out into their own module.

ndc-reference/bin/reference/main.rs
purugupta99: Updated!

ndc-reference/bin/reference/main.rs
purugupta99: Updated!

purugupta99:
> At the very least I think we should think about splitting these functions out into their own module.

This would require refactoring to move the structs and relevant functions also to another module, if that is fine I can do it.

ndc-reference/bin/reference/main.rs
purugupta99: Silly question, where do we update the text for these function explanations? ^^"

paf31:
@purugupta99 Are any of these blocking tests from being written on engine which are essential?

I think we should keep a branch for tests for now, but not merge this as-is. I will work on making the reference agent into a library which can be customized on the engine side, because I don't think the engine tests should inform what goes into the reference agent directly. On the other hand, if some specified feature is not demonstrable with the reference agent then that's an issue.

The question is - do we need to be able to demonstrate every variation of a feature (commands in this case), and is it worth the extra code? I think the library option should give us the best of both worlds, but we can keep this branch maintained in the short term.

purugupta99:
> Are any of these blocking tests from being written on engine which are essential?

Yes, I'm not able to write tests for procedures because we have `upsert_article` procedure only which needs an object as argument which is not supported currently (PR in pipeline). Also for permissions, need atleast 2 functions to apply different roles on them for tests. Rest of the tests with different input/output types can be ignored for now.

> I don't think the engine tests should inform what goes into the reference agent directly

Agreed! Will wait for the reference agent library and maintain this separately for now.

> I think we should keep a branch for tests for now

Yes currently we have a separate tests [PR](https://github.com/hasura/v3-engine/pull/58) which I run locally before merging any command related feature.

paf31:
> Also for permissions, need atleast 2 functions to apply different roles on them for tests.

Can't you use the same function twice?

> Yes currently we have a separate tests https://github.com/hasura/v3-engine/pull/58 which I run locally before merging any command related feature.

I meant more like maintaining this branch, and pointing to it from v3-engine if need be. It's preferable that we refer to a versioned tag, but if we need these functions for tests then that is an option.

purugupta99:
@Naveenaidu Why do we need rows in case of `get_article_by_id`? We should simply return the object as the value corresponding to the `__value` key.

```json
[
  {
    "rows": [
      {
        "__value": {
           "article_id": 1,
           "author_id": 1,
           "title": "The Next 700 Programming Languages"
        }
      }
    ]
  }
]
```

Naveenaidu:
> ```json
> {
>            "article_id": 1,
>            "author_id": 1,
>            "title": "The Next 700 Programming Languages"
>         }
> ```

Thanks! I have reverted this change.

purugupta99:
Closing this PR in lieu of https://github.com/hasura/v3-engine/pull/112 which introduces a custom connector agent in the `v3-engine` repo itself.

PR #27, opened 2023-08-31
Support for filtering, aggregating and ordering by nested fields

dmoverton:
Modifies the ndc spec to allow referring to a nested field of an object-valued column when filtering aggregating and ordering. This is needed for databases that support nested objects, such as MongoDB.

We add a `ColumnSelector` type which is used instead of column names in `ComparisonTarget`, `Aggregate` and `OrderByTarget`. A `ColumnSelector` may refer to a single scalar column (represented by a JSON string) or to a path within an object-valued column (represented by an array of strings). 

This work is based on a similar change we made for the v2 DC API. See https://github.com/hasura/graphql-engine-mono/pull/9940 and https://github.com/hasura/graphql-engine-mono/pull/10082.

### DIFF
--- a/ndc-client/src/models.rs
+++ b/ndc-client/src/models.rs
@@ -464,11 +464,11 @@ pub enum BinaryComparisonOperator {
 // ANCHOR_END: BinaryComparisonOperator
 
 // ANCHOR: ColumnSelector
-#[derive(Clone, Debug, PartialEq)]
 /// ColumnSelector consists of a column name (the first element) and possibly a 
 /// path to a field within a nested objects in that column.
 /// Multi-element ColumnSelectors are only valid for databases that support nested objects,
 /// e.g. MongoDB.
+#[derive(Clone, Debug, PartialEq)]
 pub struct ColumnSelector(pub Vec<String>);
 
 /// A string deserializes to a a single-element ColumnSelector.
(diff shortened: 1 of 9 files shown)


ndc-client/src/models.rs
```diff
@@ -454,20 +463,100 @@ pub enum BinaryComparisonOperator {
 }
 // ANCHOR_END: BinaryComparisonOperator
 
+// ANCHOR: ColumnSelector
+/// ColumnSelector consists of a column name (the first element) and possibly a 
+/// path to a field within a nested objects in that column.
+/// Multi-element ColumnSelectors are only valid for databases that support nested objects,
+/// e.g. MongoDB.
+#[derive(Clone, Debug, PartialEq)]
+pub struct ColumnSelector(pub Vec<String>);
```
soupi, line 472: Have you considered representing this as a `(String, Vec<String>)` instead?

dmoverton: Yeah, that might be a good idea.

manasag:
Is this PR still required?

dmoverton:
> Is this PR still required?

Yes, it was postponed until after the Mongo beta release, but I plan to get back to it next week. I suspect it will need to be reworked a bit, but we definitely want to support filtering and ordering on nested fields in MongoDB at least. I think there are some other connectors waiting for it too.

dmoverton:
This was pretty out-of-date so I'm closing it in favour of a new PR https://github.com/hasura/ndc-spec/pull/138.

PR #28, opened 2023-08-31
Procedure calling convention should match functions

paf31:
This does not include the specification yet, because the mutations section has not been written.

Later, it should include something like the following:

---

## Calling Convention

Procedures are invoked using the mutation endpoint, and return their response in the corresponding `MutationOperationResults` object.

Just like functions, procedures should return a single row and a single column, named `__value`, whose type is given by the procedure's return type.

As with functions, there is no need to define the wrapper type in the `object_types` section of the schema response.

### DIFF
--- a/ndc-reference/bin/reference/main.rs
+++ b/ndc-reference/bin/reference/main.rs
@@ -1499,12 +1499,17 @@ async fn execute_mutation_operation(
                                 );
                             }
                         }
-                        Ok(vec![row])
+                        Ok(row)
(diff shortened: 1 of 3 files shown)


daniel-chambers:
@paf31 Shouldn't the return type declared in the procedure's schema type also be changed to be an array of named articles, instead of just named article?

purugupta99 (approved):
LGTM! Thanks for the PR.

PR #29, opened 2023-09-05
Argument position

dmoverton:
Add optional argument position index to `ArgumentInfo`. This is for connectors that require arguments to be passed in a specific order, rather than by name.

### DIFF
--- a/ndc-client/src/models.rs
+++ b/ndc-client/src/models.rs
@@ -212,6 +212,8 @@ pub struct ArgumentInfo {
     /// The name of the type of this argument
     #[serde(rename = "type")]
     pub argument_type: Type,
+    /// Zero-based index position of this argument in parameter list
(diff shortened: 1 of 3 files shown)


paf31:
I'm not seeing why this is needed. Suppose you have a function `f(a, b, c)` which takes 3 positional arguments which you want to expose as an NDC function. You can publish a `FunctionInfo` with

```
{ 
  "arguments": {
    "a": ...,
    "b": ...
    "c": ...
  },
  ...
}
```

i.e. unordered arguments. When you receive a query, you will get `a`, `b` and `c` in some order. It is up to the connector to know what order they must be provided to `f`. That is part of the implementation and shouldn't be part of the specification or the schema response.

dmoverton:
We have worked out another way to do positional argument that does not require a change to `ndc-spec`.

PR #30, opened 2023-09-06
Fix the non-deterministic version of IN

paf31:
@BenoitRanque spotted an issue with the `IN` operator, which was mixing up two different uses of `Vec`: one to represent disjunctions across array relationships, and a second to represent arrays of values.

Both express disjunctions, so we can fix this by looping over all `Vec`s and looking for an equality match, which is nice and simple.

However, as @BenoitRanque points out, it's less simple for connectors which generate SQL or other query documents. There, we'd need to switch between `IN` and `ANY IN` as we look at different elements of the right hand side. For that reason, I've added a note to the specification which explicitly says that supporting non-deterministic right-hand-sides of `IN` is _not required_.

### DIFF
--- a/specification/src/specification/queries/filtering.md
+++ b/specification/src/specification/queries/filtering.md
@@ -70,7 +70,7 @@ Binary comparison operators are denoted by expressions with a `type` field of `b
 
 See type [`ComparisonValue`](../../reference/types.md#comparisonvalue) for the valid inhabitants of the `value` field.
 
-_Note_: in general, `PathElement`s in a `ComparisonValue` can refer to _array_ relationships. However, in the case of the `in` operator, such requests can be very difficult to implement, and for practical purposes, it is not very useful to express such queries. Therefore, connectors can expect `PathElements` in `ComparisonValue`s to always only refer to _object_ relationships, and fail with a `Bad Request` error otherwise.
+_Note_: in general, `PathElement`s in a `ComparisonValue` can refer to _array_ relationships. However, in the case of the `in` operator, such requests can be very difficult to implement, and for practical purposes, it is not very useful to express such queries. Therefore, in the case of the `in` operator, connectors can expect `PathElements` in `ComparisonValue`s to always only refer to _object_ relationships, and fail with a `Bad Request` error otherwise.
 
 ```json
 {{#include ../../../../ndc-reference/tests/query/predicate_with_in/request.json:1 }}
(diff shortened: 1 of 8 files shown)


specification/src/specification/queries/filtering.md
```diff
@@ -93,24 +70,11 @@ Binary comparison operators are denoted by expressions with a `type` field of `b
 
 See type [`ComparisonValue`](../../reference/types.md#comparisonvalue) for the valid inhabitants of the `value` field.
 
+_Note_: in general, `PathElement`s in a `ComparisonValue` can refer to _array_ relationships. However, in the case of the `in` operator, such requests can be very difficult to implement, and for practical purposes, it is not very useful to express such queries. Therefore, connectors can expect `PathElements` in `ComparisonValue`s to always only refer to _object_ relationships, and fail with a `Bad Request` error otherwise.
```
BenoitRanque, line 73: > Therefore, connectors can expect `PathElements` in `ComparisonValue`s to always only refer to _object_ relationships

Is this always true, or only true for the `in` operator? If always true, this would allow simplifying SQL a great deal, paths become simple left joins.

However, there's a drawback to not supporting this* at all, at least in clickhouse. In terms of usefulness, it's effectively an exists check with a comparison. However, the combination of exists + root columns comparisons cannot be implemented soundly without correlated subqueries, while paths that effectively evaluate to arrays of values are much easier to implement.

*where this is paths that traverse array relationships. I have no issues with removing right hand array relationships in paths for binary array comparison operators

specification/src/specification/queries/filtering.md
```diff
@@ -93,24 +70,11 @@ Binary comparison operators are denoted by expressions with a `type` field of `b
 
 See type [`ComparisonValue`](../../reference/types.md#comparisonvalue) for the valid inhabitants of the `value` field.
 
+_Note_: in general, `PathElement`s in a `ComparisonValue` can refer to _array_ relationships. However, in the case of the `in` operator, such requests can be very difficult to implement, and for practical purposes, it is not very useful to express such queries. Therefore, in the case of the `in` operator, connectors can expect `PathElements` in `ComparisonValue`s to always only refer to _object_ relationships, and fail with a `Bad Request` error otherwise.
```
BenoitRanque, line 73: > can expect `PathElements` in `ComparisonValue`s to always only refer to _object_ relationships, and fail with a `Bad Request` error otherwise.

Note: assuming the phrasing here is intentional, and connectors are _not_ required to fail if this happens.
If connectors are required to fail we could phrase this strongly:

> can expect `PathElements` in `ComparisonValue`s to always only refer to _object_ relationships, and **must** fail with a `Bad Request` error otherwise.

BenoitRanque (approved):
Looks good, thanks for the clarification

specification/src/specification/queries/filtering.md
paf31: Right, I left it like this so that we could put it behind a capability if we decide it's useful later.

paf31: Note though, there's no loss of expressivity, since `IN` is just an `or` expression in disguise anyway.

PR #31, opened 2023-09-06
Add ErrorResponse type and require it in the response body

paf31:
Adds a structured error type


sordina (approved):
Looks good to me. The only thing I'd recommend is actually making ErrorResponse an Enum with only one constructor with the current implementation but a 'type' discriminator so that specialised error reporting can be extended in future.

paf31:
Well adding a new type of error later would be a breaking change anyway, since our versions have to track producers and consumers, so I'm inclined to use the HTTP error code for now. If there is some new error condition that we discover that must be specially handled by clients, then we could make a major release to add that.

PR #37, opened 2023-09-14
forward auth bearer token

krushanbauva:
v3-engine [PR](https://github.com/hasura/v3-engine/pull/76)

### DIFF
--- a/ndc-client/src/apis/default_api.rs
+++ b/ndc-client/src/apis/default_api.rs
@@ -62,10 +62,7 @@ pub async fn capabilities_get(
             }
 
             if let Some(ref bearer_token) = configuration.bearer_access_token {
-                req_builder = req_builder.header(
-                    reqwest::header::AUTHORIZATION,
-                    "Bearer ".to_owned() + bearer_token.as_str(),
-                );
+                req_builder = req_builder.bearer_auth(bearer_token.as_str());
             }
 
             let req = req_builder.build()?;
@@ -117,10 +114,7 @@ pub async fn explain_post(
             }
 
             if let Some(ref bearer_token) = configuration.bearer_access_token {
-                req_builder = req_builder.header(
-                    reqwest::header::AUTHORIZATION,
-                    "Bearer ".to_owned() + bearer_token.as_str(),
-                );
+                req_builder = req_builder.bearer_auth(bearer_token.as_str());
             }
 
             req_builder = req_builder.json(&query_request);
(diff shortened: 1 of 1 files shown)


paf31 (dismissed):

paf31:
@sordina Hopefully this works with what you have planned for hub too, can you please take a look?

@krushanbauva This looks okay to me but I can't test it easily. If we can do an end-to-end test with a hub connector then I can merge this tomorrow. Or feel free to merge it once you've done such a test.

ndc-client/src/apis/default_api.rs
```diff
@@ -61,6 +61,13 @@ pub async fn capabilities_get(
                     .header(reqwest::header::USER_AGENT, user_agent.clone());
             }
 
+            if let Some(ref bearer_token) = configuration.bearer_access_token {
+                req_builder = req_builder.header(
+                    reqwest::header::AUTHORIZATION,
+                    "Bearer ".to_owned() + bearer_token.as_str(),
+                );
```
paf31, line 68: Consider

```suggestion
                req_builder = req_builder.bearer_auth(bearer_token.as_str());
```

See https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.bearer_auth

krushanbauva:
I have tested this with the reference agent and it is working as expected (i.e. it is sending the configured `Authorization` header)
@sordina I will need some help if we want to test this with a hub connector.

krushanbauva:
I was able to test it out end to end locally using a sendgrid connector with the help of @sordina and it works as expected.
The only thing that's remaining is when the request is `unauthorized`, it fails to deserialize the response into a JSON value [here](https://github.com/hasura/ndc-spec/blob/fde4c88b7f0bd02d0bea9ec537991079930cdb01/ndc-client/src/apis/default_api.rs#L173) and hence the error reporting is not accurate.
Currently, when the request is unauthorized, the error we get is
```json
{
  "data": null,
  "errors": [
    {
      "message": "internal communication error: request to connector failed",
      "path": null
    }
  ]
}
```
This can be improved to indicate that the request is unauthorized but I think this needs to be done on the connector side so it sends a valid JSON.

paf31 (approved):

paf31:
> This can be improved to indicate that the request is unauthorized but I think this needs to be done on the connector side so it sends a valid JSON response in case of unauthorized request.

Right, I think it's this issue: https://github.com/hasura/ndc-hub/pull/27/files#r1327530869

PR #38, opened 2023-09-14
Don't generate test cases with == null in predicates

paf31:


ndc-test/src/lib.rs
```diff
@@ -504,6 +504,7 @@ async fn test_select_top_n_rows_with_predicate<C: Connector>(
             variables: None,
         };
 
+        println!("{:?}", query_request);
```
danieljharvey, line 507: Do we still want this here?

danieljharvey (dismissed):

ndc-test/src/lib.rs
paf31: 🤦

paf31:
Ugh, sorry @danieljharvey, the rules are set up to discard stale reviews. Do you mind hitting OK again?

danieljharvey (approved):

PR #39, opened 2023-09-14
Use colored library in ndc-test, extract report into own function

paf31:
@sordina This should make it possible to reduce code in hub, and turn off color manually if needed.


danieljharvey (approved):
Nice!

PR #41, opened 2023-09-25
forward `headers` to connector

krushanbauva:
This PR extends struct `Configuration` to include a new optional field `headers`.
These headers are simply forwarded to the data connectors while building request.


paf31:
What about using `default_headers` like I linked?

krushanbauva:
> What about using `default_headers` like I linked? The point being, I don't think we need a new piece of configuration, since the `Client` is already there.

Yes, that makes sense.

PR #43, opened 2023-10-05
Snapshot testing in ndc-test

paf31:
This adds a `--snapshots-dir` argument to `ndc-test` which writes any generated request/response pairs to snapshots on disk, and checks responses against any snapshots which are already on disk.


ndc-test/src/lib.rs
```diff
@@ -121,9 +144,59 @@ async fn nest<A, F: Future<Output = A>>(name: &str, results: &RefCell<TestResult
     result
 }
 
+// fn snapshot_test_with_current_path<R>(
```
paf31, line 147: TODO: remove this

PR #45, opened 2023-10-10
Switch all encodings to camelCase

paf31:


ndc-reference/tests/query/table_argument_order_by/expected.json
```diff
@@ -2,10 +2,10 @@
   {
     "rows": [
       {
-        "id": 2
+        "id": 1
       },
       {
-        "id": 1
+        "id": 2
```
sordina, line 8: Is this relevant to the casing change?

sordina (changes requested):
@paf31 This looks good just two concerns:

* Can we be sure that we've covered all casing changes?
* Is this intended to be out for the October releases? What will the downstream adoption look like? Could we have an overview of what the subsequent changes downstream will be required in the PR description?

ndc-reference/tests/query/table_argument_order_by/expected.json
paf31: No, good catch. I'll look into this, thanks.

PR #49, opened 2023-10-31
use empty object instead of json value for leaf capabilities

abhinav-hasura:
This conveys the intent of the field more clearly and also avoids accepting non-object values

### DIFF
--- a/ndc-client/src/models.rs
+++ b/ndc-client/src/models.rs
@@ -32,7 +32,7 @@ pub struct CapabilitiesResponse {
 /// This is an empty struct to allow for future sub-capabilities.
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
 pub struct LeafCapability {}
-// ANCHOR_END: Capabilities
+// ANCHOR_END: LeafCapability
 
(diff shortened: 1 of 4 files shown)


paf31 (approved):
Thanks!

abhinav-hasura:
Sorry, just pushed another minor commit, can you please re-approve?

ndc-client/src/models.rs
```diff
@@ -27,16 +27,23 @@ pub struct CapabilitiesResponse {
 }
 // ANCHOR_END: CapabilitiesResponse
 
+// ANCHOR: LeafCapability
+/// A unit value to indicate a particular leaf capability is supported.
+/// This is an empty struct to allow for future sub-capabilities.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+pub struct LeafCapability {}
```
paf31, line 34: We should also update the `mdbook` to include this type in `types.md` please.

abhinav-hasura: Done

paf31 (approved):

PR #53, opened 2023-11-03
remove auth and other headers from api configuration

paritosh-08:
This PR removes auth and other headers from the API Configuration struct. This PR also adds some documentation around how we can add headers to be sent in API requests.

### DIFF
--- a/ndc-client/src/apis/configuration.rs
+++ b/ndc-client/src/apis/configuration.rs
@@ -1,3 +1,5 @@
+use reqwest::header::{HeaderMap, HeaderValue};
+
 #[derive(Debug, Clone)]
 pub struct Configuration {
     pub base_path: String,
@@ -6,6 +8,8 @@ pub struct Configuration {
     pub basic_auth: Option<BasicAuth>,
     pub oauth_access_token: Option<String>,
     pub bearer_access_token: Option<String>,
+    // TODO: should we remove other access tokens and basic_auth?
+    pub headers: HeaderMap<HeaderValue>,
     pub api_key: Option<ApiKey>,
 }
 
--- a/ndc-client/src/apis/configuration.rs
+++ b/ndc-client/src/apis/configuration.rs
@@ -1,4 +1,4 @@
-/// Configuration for the API.
+/// Configuration for the API client
 /// Contains all the information necessary to perform requests.
 /// 
 /// Please note that there is no headers field, headers (for authentication and custom headers) are supposed to be added
@@ -37,13 +37,4 @@ pub struct Configuration {
     pub base_path: String,
     pub user_agent: Option<String>,
     pub client: reqwest::Client,
-    pub api_key: Option<ApiKey>,
-}
-
-pub type BasicAuth = (String, Option<String>);
-
-#[derive(Debug, Clone)]
-pub struct ApiKey {
-    pub prefix: Option<String>,
-    pub key: String,
 }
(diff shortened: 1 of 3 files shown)


ndc-client/src/apis/configuration.rs
```diff
@@ -1,11 +1,14 @@
+use reqwest::header::{HeaderMap, HeaderValue};
+
 #[derive(Debug, Clone)]
 pub struct Configuration {
     pub base_path: String,
     pub user_agent: Option<String>,
     pub client: reqwest::Client,
```
paf31, line 7: `client` provides a way to edit the default set of headers. Why not use that?

ndc-client/src/apis/configuration.rs
```diff
@@ -1,11 +1,14 @@
...
     pub basic_auth: Option<BasicAuth>,
     pub oauth_access_token: Option<String>,
-    pub bearer_access_token: Option<String>,
+    // TODO: should we remove oauth access tokens and basic_auth as well?
```
paf31, line 10: Yes I think so.

ndc-client/src/apis/configuration.rs
paritosh-08: Are you talking about [`ClientBuilder`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html)?

So, the Configuration will have the client (already configured with default headers). While initializing, the user will have to make sure that the appropriate headers are already part of the client.

I am okay with this change, the only downside is that while initializing, we will have to create a [`ClientBuilder`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html) first, then [add default headers](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.default_headers) and then create client (and handle errors as the [`ClientBuilder.build`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.build) returns `Result<Client>`).

ndc-client/src/apis/configuration.rs
paritosh-08: Done

paf31 (approved):

paritosh-08:
@paf31 I have removed the headers.

I have also added some documentation around how the client is supposed to be built with some default set of headers. Can you please review that as well?

ndc-client/src/apis/configuration.rs
```diff
@@ -1,11 +1,42 @@
+/// Configuration for the API.
```
paf31, line 1:
```suggestion
/// Configuration for the API client
```

ndc-client/src/apis/configuration.rs
```diff
@@ -1,11 +1,42 @@
...
+/// 
 #[derive(Debug, Clone)]
 pub struct Configuration {
     pub base_path: String,
     pub user_agent: Option<String>,
     pub client: reqwest::Client,
-    pub basic_auth: Option<BasicAuth>,
-    pub oauth_access_token: Option<String>,
-    pub bearer_access_token: Option<String>,
     pub api_key: Option<ApiKey>,
```
paf31, line 40: Let's remove this one too while we're at it, please.

paf31 (approved):
Looks good, please feel free to merge after removing `api_key` and `ApiKey`.

ndc-client/src/apis/configuration.rs
paritosh-08: Thanks, done

ndc-client/src/apis/configuration.rs
paritosh-08: Done

paf31 (approved):
Thanks!

PR #60, opened 2023-11-09
[DRAFT] Mutations

paf31:


ndc-client/src/models.rs
```diff
@@ -613,6 +717,27 @@ pub enum MutationOperation {
 }
 // ANCHOR_END: MutationOperation
 
+// ANCHOR: RowUpdate
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(tag = "type", rename_all = "snake_case")]
+#[schemars(title = "Row Update")]
+pub enum RowUpdate {
+    CustomOperator {
```
danieljharvey, line 725: Is there a potential third category here, used for stuff like `inc`? Where we wouldn't need a new value?

ndc-client/src/models.rs
```diff
@@ -60,6 +61,17 @@ pub struct QueryCapabilities {
 }
 // ANCHOR_END: QueryCapabilities
 
+// ANCHOR: MutationCapabilities
+#[skip_serializing_none]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Mutation Capabilities")]
+pub struct MutationCapabilities {
+    /// Whether or not nested inserts to related collections are supported
+    pub nested_inserts: Option<LeafCapability>,
```
danieljharvey, line 70: Should we consider leaving all nested stuff out of the first iteration of mutations entirely, with an eye on looking to a nicer way to compose smaller operations together?

ndc-client/src/models.rs
```diff
@@ -163,6 +177,16 @@ pub struct AggregateFunctionDefinition {
 }
 // ANCHOR_END: AggregateFunctionDefinition
 
+// ANCHOR: UpdateOperatorDefinition
+/// The definition of an update operator on a scalar type
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Update Operator Definition")]
+pub struct UpdateOperatorDefinition {
+    /// The type of the argument to this operator
+    pub argument_type: Type,
```
danieljharvey, line 186: Thinking again of something like an `increment` operator, does making this `Type` optional make sense? Or there is nicer way to express this?

danieljharvey: Ignore this, I misunderstood - this is the type of the target column, ie `int`.

ndc-client/src/models.rs
```diff
@@ -589,19 +619,93 @@ pub struct ExplainResponse {
...
+    Update {
+        post_update_check: Option<Expression>,
+        /// The fields to return for the rows affected by this update operation
+        returning_fields: Option<IndexMap<String, Field>>,
+        /// The name of a collection
+        collection: String,
+        /// The updates to make to the matched rows in the collection
+        updates: Vec<RowUpdate>,
+        #[serde(rename = "where")]
+        r#where: Option<Expression>,
```
danieljharvey, line 707: nit: we should probably call this field `predicate` to match the `Delete` case (assuming these are the same thing?)

danieljharvey:
This all seems sensible, I think it would be good to come to an agreement on the scope size of the first iteration of this. I would argue we could leave out nested stuff as a starting point, but happy to be convinced otherwise.

PR #61, opened 2023-11-09
Revisions to initial capabilities structure

paf31:
Some breaking changes to capabilities before 0.1.0:

- Add an `aggregates` capability
- Move `relation_comparisons` and `order_by_aggregate` under relationship, since they require it
- Make `query` non-optional (it will only exist to contain `aggregates` and `variables` and group together other query-related things
- Rename `foreach` to `variables`

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ cargo build
 cargo test
 ```
 
-### Run the reference agent
+### Run the reference connector
 
 ```sh
 (cd ndc-reference; cargo run)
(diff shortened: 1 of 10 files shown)


daniel-chambers:
If we're adding aggregates as a query capability, is it also worth adding the other capabilities of a query, such as pagination (offset and limit) and ordering?

danieljharvey:
This all seems completely reasonable.

paf31:
I can make a follow-up PR for the ordering capability, because it needs changes to the test runner, but I think we need pagination in order to be able to generate even the most basic test cases in the test runner (without doing a `SELECT *`), so I'm less sure about that one.

codedmart (approved):

PR #63, opened 2023-11-13
Upgrade indexmap to v2.

SamirTalwar:
Let's bump this before stabilizing. It means all consumers can use indexmap v2 as well.

indexmap v1 is still used internally by a couple of dependencies, but hopefully those will fade away over time. They do not pollute the public API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -647,9 +647,9 @@ checksum = "8a9ee70c43aaf417c914396645a0fa852624801b24ebb7ae78fe8272889ac888"
 
 [[package]]
 name = "hashbrown"
-version = "0.14.0"
+version = "0.14.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


paf31:
@abhinav-hasura Good on the engine side?

abhinav-hasura:
> @abhinav-hasura Good on the engine side?

Should be good, assuming no compilation issues

paf31 (approved):

SamirTalwar:
Looks like the engine is not up-to-date with the latest ndc-spec and so suffers from many compilation issues on upgrading. I'm afraid I don't have the time to go through all that.

I'll leave it to whoever is responsible for keeping the engine synced with model changes.

paf31:
> Looks like the engine is not up-to-date with the latest ndc-spec and so suffers from many compilation issues on upgrading. I'm afraid I don't have the time to go through all that.

@abhinav-hasura Can you please try to say how much work it would be to update engine to indexmap-v2?

PR #64, opened 2023-11-14
linting

plcplc:
Just fixing a few typos and dead links.

### DIFF
--- a/specification/src/tutorial/schema.md
+++ b/specification/src/tutorial/schema.md
@@ -1,6 +1,6 @@
 # Schema
 
-The [schema endpoint](../specification/schema/index.html) should return data describing the data connector's scalar and object types, along with any collections, functions and procedures which are exposed.
+The schema endpoint should return data describing the data connector's scalar and object types, along with any collections, functions and procedures which are exposed.
 
 ```rust,no_run,noplayground
 {{#include ../../../ndc-reference/bin/reference/main.rs:schema1}}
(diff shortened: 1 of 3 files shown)


specification/src/tutorial/schema.md
```diff
@@ -1,6 +1,6 @@
 # Schema
 
-The [schema endpoint](../specification/schema/README.md) should return data describing the data connector's scalar and object types, along with any collections, functions and procedures which are exposed.
+The [schema endpoint](../specification/schema/index.html) should return data describing the data connector's scalar and object types, along with any collections, functions and procedures which are exposed.
```
plcplc, line 3: I'm not sure this is the right way to link in our markdown setup (i.e., it looks like it's confusing sources and compiled outputs, but the link that ended up in the compiled html still referenced a `README.md`, which is wrong)

paf31: This looks like a known bug https://github.com/rust-lang/mdBook/issues/984.

I don't think fixing it with a link to `.html` is really better, because it will break the link in the original markdown for anyone reading on GitHub (not recommended, but still).

There is a script in the issue to work around this, but I would say let's just remove this link for now.

paf31, line 3:
```suggestion
The schema endpoint should return data describing the data connector's scalar and object types, along with any collections, functions and procedures which are exposed.
```

paf31: Merging with this for now, to unblock the other fixes here. Thanks!

paf31 (approved):

PR #65, opened 2023-11-14
Mention that aggregate function are used in sorting

plcplc:


specification/src/specification/schema/scalar-types.md
```diff
@@ -41,6 +41,8 @@ For example:
 
 Aggregation functions extend the query AST with the ability to express new aggregates within the `aggregates` portion of a query.
 
+They also allow sorting the query results via the `order_by` query field.
```
paf31, line 44:
```suggestion
Aggregation functions extend the query AST with the ability to express new aggregates within the `aggregates` portion of a query. They also allow sorting the query results via the `order_by` query field.
```

paf31 (approved):

paf31:
Thanks!

PR #70, opened 2023-11-27
Nested field selection

dmoverton:
Add support for selecting fields of nested objects and arrays in queries.

We extend `Field::Column` with an optional `NestedField` property which allows for selection within nested arrays and fields of nested objects:

```rust
pub enum Field {
    Column {
        column: String,
        nested_field: Option<NestedField>
    },
    Relationship {
        query: Box<Query>,
        /// The name of the relationship to follow for the subquery
        relationship: String,
        /// Values to be provided to any collection arguments
        arguments: BTreeMap<String, RelationshipArgument>,
    },
}
```
If `nested_field == None` then the column is assumed to have no internal structure (i.e. it contains scalar values).
A `NestedField` can be either an object or an array:
```rust
pub enum NestedField {
    Object(NestedObject),
    Array(NestedArray)
}

pub struct NestedObject {
    pub fields: IndexMap<String, Field>,
}

pub struct NestedArray {
    pub field: Box<Option<NestedField>>,
}
```

A `NestedObject` contains a map of subselections for nested `Field`s.

A `NestedArray` just contains an optional `NestedField`.
At the moment this simply records the fact that the field represents an array value, and allows subselection
on any objects nested within the array elements.
In future we may want to extend `NestedArray` to allow other queries on arrays, e.g. filtering, ordering, limit, offset, and aggregates.


ndc-reference/universities.json
```diff
@@ -0,0 +1,2 @@
+{"id": 1, "name": "Queen Mary University of London", "location": { "city": "London", "country": "UK"}, "professors": [{"id": 1, "first_name": "Peter", "last_name": "Landin"}]}
+{"id": 2, "name": "Chalmers University of Technology", "location": { "city": "Gothenburg", "country": "Sweden"}, "professors": [{"id": 2, "first_name": "John", "last_name": "Hughes"}, {"id": 3, "first_name": "Koen", "last_name": "Claessen"}]}
```
paf31, line 2: ID 3 doesn't appear in `authors` if this is meant to act like a foreign key?

ndc-reference/bin/reference/main.rs
```diff
@@ -103,12 +128,14 @@ fn init_app_state() -> AppState {
     // Read the CSV data files
     let articles = read_csv("articles.csv").unwrap();
     let authors = read_csv("authors.csv").unwrap();
+    let universities = read_json_lines("universities.json").unwrap();
```
paf31, line 131: Can we use CSV like for everything else? Otherwise we can switch the others to JSON but I think it would be nice to have one format.

ndc-reference/bin/reference/main.rs
```diff
@@ -1477,6 +1505,81 @@ fn eval_comparison_value(
...
+                    }),
+                )
+            })?;
+            let mut row = IndexMap::new();
+            for (field_name, field) in fields.iter() {
+                row.insert(
+                    field_name,
+                    eval_field(collection_relationships, variables, state, field, &full_row)?,
+                );
+            }
```
paf31, line 1533: Is it possible to extract this common function from `execute_query` and reuse it here?

ndc-reference/tests/query/nested_array_select/request.json
```diff
@@ -0,0 +1,30 @@
...
+    "query": {
+        "fields": {
+            "id": {
+                "type": "column",
+                "column": "id"
+            },
+            "professors": {
+                "type": "column",
+                "column": "professors",
+                "nested_field": {
```
paf31, line 14: Can we call this `fields` (and the same for `array`) to be consistent everywhere?

ndc-reference/universities.json
paf31: I see why you picked this, but it'd be nice if we could come up with some data format which exercised all code paths, without repeating the data.

paf31 (commented):
Looks very nice in general, thank you for working on this.

I made a few suggested changes, and there are some bigger pieces left to work on: ndc-test, and the book, which needs a new subsection for nested fields, and a new section walking through the `eval_nested_field` function of the reference agent.

I think we should ideally leave this PR open until those are done, and in the meantime, test this extensively in engine, ideally with some real connectors too.

We probably ought to also add a capability for nested fields now.

ndc-reference/tests/query/nested_array_select/request.json
dmoverton: Done

ndc-reference/bin/reference/main.rs
dmoverton: Done

ndc-reference/bin/reference/main.rs
dmoverton: JSON is certainly more convenient for nested structures than CSV. I'll change the other files to JSON.

dmoverton: Done

ndc-client/src/models.rs
```diff
@@ -334,13 +334,42 @@ pub enum Aggregate {
...
+// ANCHOR_END: NestedField
+
 // ANCHOR: Field
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
 #[serde(tag = "type", rename_all = "snake_case")]
 #[schemars(title = "Field")]
 pub enum Field {
     Column {
         column: String,
+        fields: Option<NestedField>
```
paf31, line 372:
```suggestion
        /// When the type of the column is a (possibly-nullable) array or object,
        /// the caller can request a subset of the complete column data,
        /// by specifying fields to fetch here.
        /// If omitted, the column data will be fetched in full.
        fields: Option<NestedField>
```

We probably ought to add a section to the spec to clarify the behavior for nullable types. E.g. if `person.address` is null and the user requests `{ person { address { city } } }`, then I think we can specify that `{ person: { address : null } }` ought to be returned.

I'd also be curious to bikeshed this name a bit. It's the best name I can currently think of, but it also a bit strange that it's field_s_ plural, but there's only one structure being described. It makes sense, because the leaves of the structure are always `FieldObject`, but that might not be the case forever. In general, it's just going to be a type-directed set of instructions for fetching a subset of data. But I can't think of a better name for that. Perhaps `transform`, but that's confusing for other reasons.

ndc-reference/bin/reference/main.rs
```diff
@@ -13,9 +16,12 @@ use axum::{
 };
 
 use indexmap::IndexMap;
-use ndc_client::models::{self, LeafCapability, RelationshipCapabilities};
+use ndc_client::models::{
+    self, LeafCapability, NestedArray, NestedField, NestedObject, RelationshipCapabilities,
```
paf31, line 20: Consider referencing these qualified as `models::NestedArray` etc. just to keep the imports under control here.

ndc-reference/bin/reference/main.rs
```diff
@@ -100,15 +109,17 @@ async fn metrics_middleware<T>(
 // ANCHOR_END: metrics_middleware
 // ANCHOR: init_app_state
 fn init_app_state() -> AppState {
-    // Read the CSV data files
-    let articles = read_csv("articles.csv").unwrap();
-    let authors = read_csv("authors.csv").unwrap();
+    // Read the JSON data files
+    let articles = read_json_lines("articles.json").unwrap();
+    let authors = read_json_lines("authors.json").unwrap();
+    let universities = read_json_lines("universities.json").unwrap();
```
paf31, line 115: Complete and utter nitpick (sorry) but we could consider renaming this to `institutions` to match things like BibTeX, and because not every author is associated with a university.

paf31 (changes requested):
I think this is very close to being mergeable but there are a few remaining supporting pieces needed:

1. I think we need to figure out the ideal demo in the reference agent - `universities` has what looks like a foreign key relationship on `authors`, but it doesn't work because of missing keys. We could consider renaming `id` to `author_id` and making it nullable.
2. There needs to be a section in `specification/src/specification/queries/field-selection.md` talking about nested field selections.
3. We need to add a (very short) RFC, but I can do this right before the PR gets merged.

paf31:
Could you look at fixing the test suite? I'll work on the spec pages.

dmoverton:
> Could you look at fixing the test suite? I'll work on the spec pages.

I'll do that. Thanks.

paf31 (commented):
Looks like you forgot to check in `authors.json` and `articles.json`?

ndc-client/src/models.rs
```diff
@@ -350,13 +350,46 @@ pub enum Aggregate {
 }
 // ANCHOR_END: Aggregate
 
+// ANCHOR: NestedObject
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+#[schemars(title = "NestedObject")]
+pub struct NestedObject {
+    pub fields: IndexMap<String, Field>,
```
paf31, line 358: Sorry I just noticed this - this seems to imply that we accept `Field::Relationship` here, i.e. relationships from nested fields. Was that intended? @abhinav-hasura Is that compatible with what you're thinking on the engine side?

Also, `Field::Column` is confusing, because presumably it would not refer to a _column_, but the name of a property in the nested object?

I'm thinking we should have `NestedField` always denote a projection function (only removing or renaming things), and that `NestedObject` should always be forced to choose from the fields of the "current object".

dmoverton: I'm not sure why we wouldn't want to allow relationships from nested fields. Is there some sort of problem that creates? I have a [draft PR for the engine](https://github.com/hasura/v3-engine/pull/255) which I've been working in conjunction with this spec PR and haven't run into any issues there so far.

paf31: So I think we need to be clear about how `Field`s are meant to be evaluated in the context of a nested object. Is there a "current object" just like we have a current row? So `Column` points to the column in the current object now? 

Also, `Relationship` might be useful, but there's no way to inform the client that a field on a nested object is a good candidate for a relationship (like `foreign_keys`).

We'll pick this up in our architecture sync today and I'll drop notes here.

ndc-client/src/models.rs
```diff
@@ -350,13 +350,46 @@ pub enum Aggregate {
...
+}
+// ANCHOR_END: NestedObject
+
+// ANCHOR: NestedArray
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+#[schemars(title = "NestedArray")]
+pub struct NestedArray {
+    pub fields: Box<Option<NestedField>>,
```
paf31, line 367: What does `None` represent here?

dmoverton: `None` means that there are no nested selections for the array elements, e.g. if the elements are scalars.

dmoverton: Example:
```json
    "query": {
        "fields": {
            "departments": {
                "type":"column",
                "column": "departments",
                "fields": {
                    "type": "array"
                }
            }
        }
```
Field `departments` is an array of strings so there are no subselections within the array elements

ndc-reference/bin/reference/main.rs
```diff
@@ -1488,6 +1510,81 @@ fn eval_comparison_value(
...
+                    .map(models::RowFieldValue)
+                    .collect::<Vec<_>>(),
+                Some(field) => array
+                    .into_iter()
+                    .map(|value| {
+                        eval_nested_field(collection_relationships, variables, state, value, field)
+                    })
+                    .collect::<Result<Vec<_>>>()?,
+            };
+            Ok(models::RowFieldValue(
```
paf31, line 1573: So we have two layers of `RowFieldValue` here, is that deliberate?

paf31: Never mind, it was the `None` case that threw me off. Can we just return `array` directly and skip the reallocation?

ndc-client/src/models.rs
paf31: But then why not use `Field::Column { fields: None, ... }`?

dmoverton: We _could_ do that now, but in the future we might want to perform other operations on arrays, e.g. sorting or filtering.

rfcs/0003-nested-field-selection.md
```diff
@@ -0,0 +1,61 @@
+# Nested Field Selection
```
paf31, line 1: [Rendered](https://github.com/hasura/ndc-spec/blob/9b4c726a5daafc342a0693fb8275de5859172957/rfcs/0003-nested-field-selection.md)

cc @dmoverton

ndc-client/src/models.rs
paf31: But those could be added as new constructors to `NestedField`, I think? If we're concerned about that, then we arguably should make `NestedArray` an enum now.

rfcs/0003-nested-field-selection.md
```diff
@@ -0,0 +1,61 @@
...
+    Object(NestedObject),
+    Array(NestedArray)
+}
+
+pub struct NestedObject {
+    pub fields: IndexMap<String, Field>,
+}
+
+pub struct NestedArray {
+    pub field: Box<NestedField>,
```
paf31, line 48: @dmoverton Note, I omitted the `Option` here, until we figure that detail out.

dmoverton: To clarify my previous comment: at the moment `Column {column: "name", fields: Some(Array(NestedArray{ fields: None }))}` is equivalent to `Column {column: "name", fields: None }` so could argue that there's no point in making `NestedArray.field` optional. However, in future we probably want to extend `NestedArray` with `limit`, `offset`, `order_by` and `predicate`:
```rust
pub struct NestedArray {
    pub fields: Box<Option<NestedField>>,
    pub limit: Option<u32>,
    /// Optionally offset from the Nth result
    pub offset: Option<u32>,
    pub order_by: Option<OrderBy>,
    #[serde(rename = "where")]
    pub predicate: Option<Expression>,
}
```
(We have already done this is v2: see https://gist.github.com/dmoverton/924c069dded11c2101db54e8afd32b26)
If/when we do this we will want `NestedArray.fields` to be optional so we can have those other things without requiring an element subselection. So I guess the question is what we want to do right now? I'm happy to change it to be non-optional if you like, but I also don't have a problem with leaving it how it is.

rfcs/0003-nested-field-selection.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Purpose
+
+Field selection is useful in a variety of contexts:
+
+- Document databases contain nested structures, and more often than not, we only want to select a subset of the available data.
+- Many relational databases also have semi-structured datatypes like `json(b)`.
+- The TypeScript connector can return object and array types, and we can pass along the selection set and only return the requested data.
+
+This proposal is a non-breaking, minimal addition to the specification, which adds _selection_ for nested structures. Aadditional operations on this sort of data (filtering, sorting, aggregation, etc.) will be covered in later proposals.
```
dmoverton, line 11:
```suggestion
This proposal is a non-breaking, minimal addition to the specification, which adds _selection_ for nested structures. Additional operations on this sort of data (filtering, sorting, aggregation, etc.) will be covered in later proposals.
```

rfcs/0003-nested-field-selection.md
paf31: Ah, got it. That makes sense. Would you mind updating the note in the RFC to this effect then?

paf31: Do we need to rename it to `fields` then?

paf31: Sorry, thinking about this some more, doesn't this mean the right solution is still to remove the outer `Option`? Specifically, there is still duplication between

```rust
Column { column: "name", fields: Some(Array(NestedArray{ fields: None, limit: None, offset: None, order_by: None, predicate: None })) }
```

and

```rust
Column { column: "name", fields: None }
```

Both return the whole array. We can't require all of the inner `Option`s to be not-simultaneously-null, but we can remove the outer `Option`.

paf31:
I'll write up the spec pages tomorrow.

rfcs/0003-nested-field-selection.md
dmoverton: Technically, we could remove the outer `Option`, but then all scalar column selections would need to be
```rust
Column { column: "name", fields: Array(...) }
```
which seems a bit odd, unless we can come up with a better name for the `Array` enum. It possibly also presents some backwards compatibility issues, if we have a connector using a newer version of the spec than the engine (although maybe we don't care about that).

paf31: Ok, so ignore the comment about removing the outer `Option`, I think I was confusing myself about something.

It will be non-breaking to add the `Option` later when we implement aggregates, but at that point we will need to figure out the return type for nested aggregates anyway. So we could remove the `Option` now with a plan to add it back in a non-breaking change later if/when we figure out that return type. Or we could assume we'll figure it out and just keep it now. I'd probably prefer the first option to minimize confusion about what the extra `Option` is for.

dmoverton: @paf31 I've made the change you suggested to remove the `Option` from `NestedArray`.

paf31 (commented):
@dmoverton I just realized you don't ever add `institutions` to the schema response?

Also, one final nitpick: I think we should consider renaming `institutions.staff.fields` to `specialities` or `fields_of_study` or something else. Even though it's the more accurate name, `fields` is confusing in the spec pages because `fields` also appears in so many types now :)

dmoverton:
> @dmoverton I just realized you don't ever add `institutions` to the schema response?

I have fixed this. However, now that `institutions` is in the schema there is a failing test case: it is generating queries that try to order by object and array fields. I'll fix this on Monday.

> 
> Also, one final nitpick: I think we should consider renaming `institutions.staff.fields` to `specialities` or `fields_of_study` or something else. Even though it's the more accurate name, `fields` is confusing in the spec pages because `fields` also appears in so many types now :)

I agree. I've changed it to specialities.

paf31:
Tests should hopefully pass now.

paf31 (approved):

paf31:
Thanks @dmoverton!

PR #71, opened 2023-11-28
use `reqwest::Url` instead of `String` as base url

paritosh-08:
Instead of appending `\<path>` to base URLs, use `Url::join("<path>")`.
JIRA: https://hasurahq.atlassian.net/browse/V3API-227

### DIFF
--- a/ndc-test/src/main.rs
+++ b/ndc-test/src/main.rs
@@ -1,4 +1,4 @@
-use std::{path::PathBuf, process::exit, str::FromStr};
+use std::{path::PathBuf, process::exit};
 
 use clap::{Parser, Subcommand};
 use ndc_client::apis::configuration::Configuration;
@@ -17,7 +17,7 @@ struct Options {
 enum Commands {
     Test {
         #[arg(long, value_name = "ENDPOINT")]
-        endpoint: String,
+        endpoint: reqwest::Url,
         #[arg(long, value_name = "SEED")]
         seed: Option<String>,
         #[arg(long, value_name = "PATH")]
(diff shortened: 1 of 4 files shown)


paf31 (approved):
Much nicer, thanks!

ndc-test/src/main.rs
```diff
@@ -45,7 +45,7 @@ async fn main() {
             };
 
             let configuration = Configuration {
-                base_path: endpoint,
+                base_path: reqwest::Url::from_str(&endpoint).unwrap(),
```
paf31, line 48: Try moving this into the `Commands` type please:

```rust
#[clap(
    takes_value = true,
    parse(try_from_str)
)]
endpoint: Url,
```

paf31 (approved):

SamirTalwar:
Yo, this is a breaking change, as the URL must now end in a trailing slash if the path is non-empty.

We will need to check that this is reasonable.

paf31:
> the URL must now end in a trailing slash if the path is non-empty.

So you get an error from `reqwest` for something like `http://domain/path` now?

paritosh-08:
My bad; this is the expected behaviour by the library: https://github.com/servo/rust-url/issues/285.

Should we do something like the suggestion in the comment: https://github.com/servo/rust-url/issues/285#issuecomment-285346010 (add trailing slash if not present)?

paritosh-08:
> > the URL must now end in a trailing slash if the path is non-empty.
> 
> So you get an error from `reqwest` for something like `http://domain/path` now?

No, while joining `query` to the above url, the `URL::join()` will return something like `http://domain/query`. However, for `http://domain/path/`, the joining will return `http://domain/path/query`.

paf31:
And what if you join with `/query`?

paritosh-08:
> And what if you join with `/query`? Same issue?

Yes.

PR #73, opened 2023-12-06
fix trailing `/` bug with url join

paritosh-08:


ndc-client/src/apis/mod.rs
```diff
@@ -33,7 +48,7 @@ impl fmt::Display for Error {
             Error::Serde(e) => ("serde", e.to_string()),
             Error::Io(e) => ("IO", e.to_string()),
             Error::ConnectorError(e) => ("response", format!("status code {}", e.status)),
-            Error::InvalidConnectorUrl(e) => ("url-parse-error", e.to_string()),
+            Error::ConnectorURLError(e) => ("url-parse-error", format!("{}", e.to_string())),
```
SamirTalwar, line 51:
```suggestion
            Error::ConnectorURLError(e) => ("url-parse-error", e.to_string()),
```

ndc-client/src/apis/mod.rs
```diff
@@ -17,13 +17,28 @@ impl fmt::Display for ConnectorError {
     }
 }
 
+#[derive(Debug)]
+pub enum ConnectorURLError {
```
SamirTalwar, line 21: We could probably just use `thiserror` here.

ndc-client/src/apis/default_api.rs
```diff
@@ -326,3 +333,37 @@ mod utils {
         });
     }
 }
+
+mod test {
```
SamirTalwar, line 337:
```suggestion
#[cfg(test)]
mod tests {
```

SamirTalwar (approved):

SamirTalwar (approved):

ndc-client/src/apis/default_api.rs
```diff
@@ -34,6 +34,23 @@ fn inject_trace_context(builder: RequestBuilder) -> RequestBuilder {
...
+    if url.path_segments().map_or(false, |mut s|
+        // It is safe to unwrap here as according to documentation of Url::path_segments()
+        // > When Some is returned, the iterator always contains at least one string (which may be empty).
+        s.next_back().unwrap() != "")
+    {
+        let mut url = url.clone();
+        // No trailing slash, add it
+        url.path_segments_mut()
+            .map_err(|_| ConnectorURLError::URLCannotBeABase())?
+            .push("");
```
paf31, line 47: Why not just push `path` here and skip the extra call to `join` (and presumably allocation)?

ndc-client/src/apis/default_api.rs
```diff
@@ -326,3 +333,38 @@ mod utils {
         });
     }
 }
+
+#[cfg(test)]
+mod tests {
```
paf31, line 338: If there's really no library for this function then we should probably make one, and move the tests there, because this seems out of scope for `ndc-spec`. I think a `url-utils` crate would be a nice little Rust OSS contribution, but I feel like something ought to exist.

ndc-client/src/apis/default_api.rs
```diff
@@ -34,6 +34,23 @@ fn inject_trace_context(builder: RequestBuilder) -> RequestBuilder {
     req_builder
 }
 
+fn append_path(url: &reqwest::Url, path: &str) -> Result<reqwest::Url, ConnectorURLError> {
+    if url.path_segments().map_or(false, |mut s|
```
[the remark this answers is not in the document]
paf31: I put my recommendation on the `phil/append-path-in-url-bug` branch, so see what you think?

ndc-client/src/apis/default_api.rs
paf31: (we can do this later, of course)

ndc-client/src/apis/default_api.rs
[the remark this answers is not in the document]
paf31: Of course, I just didn't want to overwrite yours. Do you mind testing it with engine, and I'll approve and you can merge and tag when you're happy?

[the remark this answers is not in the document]
paritosh-08: Done. I can confirm that the fix works for urls with non-zero path segments with and without trailing slash.

PR #74, opened 2023-12-07
If the server responds with an error and no body, report the error.

SamirTalwar:
A decoding error doesn't really help here.

### DIFF
--- a/ndc-client/src/apis/default_api.rs
+++ b/ndc-client/src/apis/default_api.rs
@@ -3,7 +3,6 @@ use opentelemetry::{
     trace::{FutureExt, Tracer},
     Context,
 };
-use reqwest::{self, RequestBuilder};
 use serde_json as json;
 use std::collections::HashMap;
 
@@ -21,7 +20,7 @@ impl ToHeaderString for HashMap<String, json::Value> {
     }
 }
 
-fn inject_trace_context(builder: RequestBuilder) -> RequestBuilder {
+fn inject_trace_context(builder: reqwest::RequestBuilder) -> reqwest::RequestBuilder {
     let ctx = Context::current();
     let mut trace_headers = HashMap::new();
     global::get_text_map_propagator(|propagator| {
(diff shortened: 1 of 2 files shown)


ndc-client/src/apis/default_api.rs
```diff
@@ -271,18 +246,26 @@ pub async fn schema_get(
...
-                    error_response,
-                };
-                Err(Error::ConnectorError(connector_error))
+                handle_error(response_status, response_content)
             }
         })
         .await
 }
 
+fn handle_error<T>(
```
paf31, line 255: We don't need the generic `T` here, since we always return `Err` anyway, so we can just return `Error` and move the `Err` wrapper to the call-site.

ndc-client/src/apis/mod.rs
```diff
@@ -4,15 +4,19 @@ use std::fmt;
 #[derive(Debug, Clone)]
 pub struct ConnectorError {
     pub status: reqwest::StatusCode,
-    pub error_response: crate::models::ErrorResponse,
+    pub error_response: Option<crate::models::ErrorResponse>,
```
paf31, line 7: The spec does require that the connector return an `ErrorResponse`, so I think we should call this case out as a new top-level constructor, so that it's more obvious on the engine side what ought to be done with it (we should capture the original status code, but report this as a bug on the connector ideally).

SamirTalwar: This would mean converting `ErrorResponse` to an `enum`. Are you OK with that?

ndc-client/src/apis/default_api.rs
SamirTalwar: Done.

soupi (dismissed):

ndc-client/src/apis/mod.rs
paf31: I still think the key is that the caller needs to be able to identify this special case and report it (without having to case on strings).  I don't think `ErrorResponse` needs to be an enum, you could just add a new alternative to `ndc_client::Error`, called `InvalidErrorResponse` or something similar, which captures the useful parts of the response object (or maybe all of it).

SamirTalwar: Good idea. I pushed a change.

ndc-client/src/apis/mod.rs
```diff
@@ -17,12 +17,29 @@ impl fmt::Display for ConnectorError {
...
+pub struct InvalidConnectorError {
+    pub status: reqwest::StatusCode,
+    pub content: serde_json::Value,
+}
+
+impl fmt::Display for InvalidConnectorError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "ConnectorError {{ status: {0}, content: {1} }}",
```
paf31, line 30:
```suggestion
            "InvalidConnectorError {{ status: {0}, content: {1} }}",
```

paf31 (approved):
Looks good, thanks!

PR #79, opened 2023-12-13
Move equality and IN operators into the operator definition enum

paf31:
### Problem

There is a single notion of equality on scalar types right now. There are two issues:

- `v3-engine` must reserve a name for the `equals` operator, and another for `in`. It would be better if the name could be separated from the operator meaning.
- There is no good way to express zero, or two-or-more equality or set-membership operators. If we relax the notion of equality to an equivalence relation, then there are several good operators, and in many cases, none are canonical. For example, a string type might have many different equalities, depending on things like collation or case-sensitivity. None are more correct than the others.

### Solution

The _name_ of an operator will be separated from its _meaning_. The meaning of an operator will be specified in its definition object in the schema, and it can be `equal`, `in` or `custom`. We can add new standard types over time without a breaking change. 

The argument type for standard operators will be determined based on the type of the column, and specified explicitly in the specification.

This allows connectors to give custom equality operators, with custom names.


ndc-client/src/models.rs
```diff
@@ -156,10 +156,15 @@ pub enum Type {
...
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(tag = "type", rename_all = "snake_case")]
 #[schemars(title = "Comparison Operator Definition")]
-pub struct ComparisonOperatorDefinition {
-    /// The type of the argument to this operator
-    pub argument_type: Type,
+pub enum ComparisonOperatorDefinition {
+    Equal,
+    In,
+    Custom {
```
abhinav-hasura, line 164: Should we also include IsNull here?

paf31: Yes I was considering that too. We can add that to the TODO list if the basic approach seems okay.

abhinav-hasura: Great - yeah the approach looks good to me!

abhinav-hasura: Actually, `IsNull` wouldn't be advertised for a particular scalar type - it applies to all columns that are nullable (maybe a capability).

But would we want other custom Unary comparison operators? (eg: IsPositive on an integer)

paf31: Tracking what we discussed separately - let's keep `IsNull` as a special unary operator, and we can add custom unary operators if the need arises.

specification/src/specification/schema/scalar-types.md
```diff
@@ -37,6 +36,26 @@ For example:
 }
 ```
 
+### Standard Comparison Operators
+
+#### `Equal`
+
+An operator defined using type `equal` tests if a column value is equal to a scalar value, another column value, or a variable.
+
+The precise semantics of this equality operator are not specified, but the operator should define an equivalence relation. That is, its argument type should be the same scalar type for which it is defined, and at the level of values, it should be reflexive, symmetric and transitive.
```
abhinav-hasura, line 45: How about specifying some more precise semantics? So that in the engine we can rely on using the equality operator from either connector in case of remote joins?

coco98: Maybe we can list out all the use-cases where we expect this to be used by the engine? And that implicitly defines the semantics?

paf31: > How about specifying some more precise semantics? So that in the engine we can rely on using the equality operator from either connector in case of remote joins?

The problem is that it's very difficult in general. You might have two completely unrelated types which happen to serialize to the same JSON representation, like `citext` and `varchar`, but with different equalities. So we can't just move the evaluation from one connector to the other. For local relationships, we can check that the types match, but we can't for remote relationships.

paf31: As discussed, we will leave this as-is, and handle this on the engine side, or with a breaking change here, when we come to implement more advanced remote joins.

codedmart (approved):

PR #80, opened 2024-01-10
Add a capability for transactional mutations

soupi:
This PR encodes the support a connector offers wrt mutations in types.
It introduces a new capability, `mutation`, to signal whether the connector supports mutations, and a field in `MutationCapabilities`, `transactional`, to signal whether the connector support running multiple mutations in a single transaction.

Additional discussion on Slack: https://hasurahq.slack.com/archives/C065TFX2V2Q/p1702996285519119

### DIFF
--- a/ndc-client/src/models.rs
+++ b/ndc-client/src/models.rs
@@ -42,7 +42,7 @@ pub struct LeafCapability {}
 pub struct Capabilities {
     pub query: QueryCapabilities,
     pub explain: Option<LeafCapability>,
-    pub mutation: Option<MutationCapabilities>,
+    pub mutation: MutationCapabilities,
     pub relationships: Option<RelationshipCapabilities>,
 }
 // ANCHOR_END: Capabilities
--- a/ndc-client/src/models.rs
+++ b/ndc-client/src/models.rs
@@ -42,6 +42,7 @@ pub struct LeafCapability {}
 pub struct Capabilities {
     pub query: QueryCapabilities,
     pub explain: Option<LeafCapability>,
+    pub mutation: Option<MutationCapabilities>,
     pub relationships: Option<RelationshipCapabilities>,
 }
 // ANCHOR_END: Capabilities
@@ -57,6 +58,16 @@ pub struct QueryCapabilities {
     pub variables: Option<LeafCapability>,
 }
 // ANCHOR_END: QueryCapabilities
+
+// ANCHOR: MutationCapabilities
+#[skip_serializing_none]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Mutation Capabilities")]
+pub struct MutationCapabilities {
+    /// Does the connector support executing multiple mutations in a transaction.
+    pub transactional: Option<LeafCapability>,
+}
+// ANCHOR_END: MutationCapabilities
 
 // ANCHOR: RelationshipCapabilities
 #[skip_serializing_none]
--- a/specification/src/specification/capabilities.md
+++ b/specification/src/specification/capabilities.md
@@ -22,8 +22,10 @@ See [`CapabilitiesResponse`](../reference/types.md#capabilitiesresponse)
 
 | Name | Description |
 |------|-------------|
-| `versions` | A [semantic versioning](https://semver.org) range of API versions which the data connector 
+| `versions` | A [semantic versioning](https://semver.org) range of API versions which the data connector
 | `capabilities.explain` | Whether the data connector is capable of describing query plans |claims to implement |
+| `capabilities.mutation` | Whether the data connector is capable of executing mutations |claims to implement |
+| `capabilities.mutation.transactional` | Whether the data connector is capable of executing multiple mutations in a transaction |claims to implement |
 | `capabilities.query.aggregates` | Whether the data connector supports [aggregate queries](queries/aggregates.md) |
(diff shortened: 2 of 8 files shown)


ndc-client/src/models.rs
```diff
@@ -42,6 +42,7 @@ pub struct LeafCapability {}
 pub struct Capabilities {
     pub query: QueryCapabilities,
     pub explain: Option<LeafCapability>,
+    pub mutation: Option<MutationCapabilities>,
```
paf31, line 45: Right now there's no notion of not supporting mutations at all. You would simply not have any procedures. So this `Option` only exists for backwards compatibility, right? If so, I'd say let's just make a breaking change now.

ndc-client/src/models.rs
```diff
@@ -58,6 +59,16 @@ pub struct QueryCapabilities {
 }
 // ANCHOR_END: QueryCapabilities
 
+// ANCHOR: MutationCapabilities
+#[skip_serializing_none]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Mutation Capabilities")]
+pub struct MutationCapabilities {
+    /// Does the connector support executing multiple mutations in a transaction.
+    pub transactional: Option<LeafCapability>,
```
paf31, line 68: I think we need to be clear about what this implies for the semantics of mutations, and I don't know that we can really do that. Perhaps we could do it with respect to reads (a mutation is transactional if its components can only be observed together as a unit, or not at all, where observed means by probing the data source using the query API).

But how will engine use this capability? The abstraction has to be useful from the consumer point of view as well, and I'm not seeing what this would enable or disable on the engine side.

ndc-client/src/models.rs
soupi: Alright. I removed the Option.

ndc-client/src/models.rs
soupi: We discussed this on slack, the tl;dr is that this would enable the engine to know whether it should make a multiple requests with one mutation in each request, or a single request containing multiple mutations, when it receives a request with multiple mutations from the user.

The difference between the two is that with multiple mutations per request once we encounter an error we terminate and return the error and no data (and it seems that rolling back is the only reasonable semantics we can thinking of atm), and with single mutations per request we do not terminate, but return both the data and errors for their respective mutations.

How should we proceed with this PR? Do we want to reword these ideas and flags?

paf31:
@0x777 Could you please add your notes regarding the GraphQL/engine side of things here?

specification/src/specification/capabilities.md
```diff
@@ -22,8 +22,10 @@ See [`CapabilitiesResponse`](../reference/types.md#capabilitiesresponse)
 
 | Name | Description |
 |------|-------------|
-| `versions` | A [semantic versioning](https://semver.org) range of API versions which the data connector 
+| `versions` | A [semantic versioning](https://semver.org) range of API versions which the data connector
 | `capabilities.explain` | Whether the data connector is capable of describing query plans |claims to implement |
+| `capabilities.mutation` | Whether the data connector is capable of executing mutations |claims to implement |
```
paf31, line 27: I think we can delete this line now that it's required anyway.

specification/src/specification/capabilities.md
```diff
@@ -22,8 +22,10 @@ See [`CapabilitiesResponse`](../reference/types.md#capabilitiesresponse)
...
+| `capabilities.mutation.transactional` | Whether the data connector is capable of executing multiple mutations in a transaction |claims to implement |
```
paf31, line 28: The suffix `claims to implement |` looks like a copy/paste typo from an older PR. Could you please fix it here?

specification/src/specification/mutations/README.md
```diff
@@ -1,6 +1,6 @@
 # Mutations
 
-The mutation endpoint accepts a mutation request, containing a collection of mutation operations to be performed in the context the data source, and returns a response containing a result for each operation.
+The mutation endpoint accepts a mutation request, containing a collection of mutation operations to be performed transactionally in the context the data source, and returns a response containing a result for each operation.
```
paf31, line 3: Can you please add a subsection below, under `Mutation Operations`, with something like the following

```md
### Multiple Operations

If the `mutations.transactions` capability is enabled, then the caller may provide multiple operations in a single request. Otherwise, the caller must provide exactly one operation.

The intent is that multiple operations ought to be performed together in a single transaction. That is, they should all succeed, or all fail together. If any operation fails, then a single `ErrorResponse` should capture the failure, and none of the operations should effect any changes to the data source.
```

specification/src/specification/capabilities.md
soupi: done.

specification/src/specification/capabilities.md
soupi: done.

specification/src/specification/mutations/README.md
soupi: done.

ndc-client/src/models.rs
paf31: Resolved previously - we will capture this in the specification with a note about error handling. It makes sense to call this "transactional", because we need to specify something along the lines of "all or none will be performed".

paf31 (approved):
Looks great, thanks! Once I figure out the release plan, I'll merge this.

PR #82, opened 2024-01-23
RFC: versioning

paf31:
[Rendered](https://github.com/hasura/ndc-spec/blob/06838b8d3601addc494f5df9b30cdff81a586f38/rfcs/0001-versioning.md)


rfcs/0001-versioning.md
```diff
@@ -0,0 +1,23 @@
...
+  - Note: previously, we specified a version range, but there is little point claiming compatibility with an older version, once support for a newer version is announced; in addition, non-contiguous version ranges do not make sense, so it makes sense to identify the single most-recent compatible version.
+  - For clarity: if a connector claims compatibility with a particular version, then any client should be able to use this connector, as long as the client is using a specification version which is compatible with the connector specification version, according to the semantic versioning specification.
+- A client can depend on a particular specification version by either:
+  - using the Rust client library provided in this repository, at the same _Git revision_.
+    - Note: the semantic versioning of the reposity tracks the _specification version_, and not the Rust library versions. While we will try to not introduce unnecessary breaking changes between major specification revisions, we can't guarantee that, and the user should be careful to depend on these Rust libraries at the correct version.
```
daniel-chambers, line 18:
```suggestion
    - Note: the semantic versioning of the repository tracks the _specification version_, and not the Rust library versions. While we will try to not introduce unnecessary breaking changes between major specification revisions, we can't guarantee that, and the user should be careful to depend on these Rust libraries at the correct version.
```

daniel-chambers, line 18: So does this imply that if we fix bugs in the Rust code that do not affect the specification, we'd update the code and move the tag to the latest commit?

paf31: Thanks!

codedmart (dismissed):

codedmart (approved):

PR #83, opened 2024-01-24
RFC: boolean expression types

paf31:
WIP

### DIFF
--- a/ndc-client/src/models.rs
+++ b/ndc-client/src/models.rs
@@ -150,6 +150,11 @@ pub enum Type {
         /// The type of the elements of the array
         element_type: Box<Type>,
     },
+    /// A predicate type for a given collection
+    Predicate {
+        /// The collection name
+        collection_name: String,
+    }
 }
 // ANCHOR_END: Type
 
--- a/ndc-client/tests/json_schema/schema_response.jsonschema
+++ b/ndc-client/tests/json_schema/schema_response.jsonschema
@@ -161,6 +161,26 @@
               ]
             }
           }
+        },
+        {
+          "description": "A predicate type for a given collection",
+          "type": "object",
+          "required": [
+            "collection_name",
+            "type"
+          ],
+          "properties": {
+            "type": {
+              "type": "string",
+              "enum": [
+                "predicate"
+              ]
+            },
+            "collection_name": {
+              "description": "The collection name",
+              "type": "string"
+            }
+          }
         }
       ]
     },
--- a/ndc-reference/bin/reference/main.rs
+++ b/ndc-reference/bin/reference/main.rs
@@ -373,8 +373,28 @@ async fn get_schema() -> Json<models::SchemaResponse> {
         },
     };
     // ANCHOR_END: schema_procedure_upsert_article
+    // ANCHOR: schema_procedure_delete_articles
+    let delete_articles = models::ProcedureInfo {
+        name: "delete_articles".into(),
+        description: Some("Delete articles which match a predicate".into()),
+        arguments: BTreeMap::from_iter([(
+            "where".into(),
+            models::ArgumentInfo {
+                description: Some("The predicate".into()),
+                argument_type: models::Type::Predicate {
+                    collection_name: "articles".into(),
+                },
+            },
+        )]),
+        result_type: models::Type::Array {
+            element_type: Box::new(models::Type::Named {
+                name: "article".into(),
+            }),
+        },
+    };
+    // ANCHOR_END: schema_procedure_delete_article
     // ANCHOR: schema_procedures
-    let procedures = vec![upsert_article];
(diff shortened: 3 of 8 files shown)


ndc-client/src/models.rs
```diff
@@ -150,6 +150,11 @@ pub enum Type {
         /// The type of the elements of the array
         element_type: Box<Type>,
     },
+    /// A predicate type for a given collection
```
soupi, line 153: Should we specify that a value of this type needs to parse into the `Expression` type?

paf31: Yes, when I write the spec pages, this will be mentioned there.

soupi (dismissed):
lgtm

paf31:
@abhinav-hasura I want to include your other proposal here for comparison. Please let me know if this is correct?

---

Instead of (or in addition to) boolean expression types, we'd add a separate `MutationOperation::ProcedureOfModel`, which would be tied to a model, and whose result type would be required to be an array of the object type associated with that model. It would get an implicit argument whose type would effectively be that model's predicate type (although that no longer _needs_ to be a first-class part of the type language now). This allows us to express things like "delete where", and even most (all?) post-conditions on inserts and updates (because the check constraint would be made against the tentative state of the collection in the data source before the transaction closed, so it makes sense to express that in terms of the collection type, even if it includes things like computed fields, which wouldn't be part of the input argument type).

soupi:
> This allows us to express things like "delete where"

Can't you already express things like "delete where" with regular boolean expressions? If your `delete_article` procedure expects an input of type `Expression { collection_name: "articles" }`, and it uses that as the where argument, isn't that effectively the same?

paf31:
@soupi Yes, it's possible in both approaches. This was an alternative design - the only point I was making was that they're not mutually incompatible, and we could potentially do both.

abhinav-hasura:
> @abhinav-hasura I want to include your other proposal here for comparison. Please let me know if this is correct?
> 
> Instead of (or in addition to) boolean expression types, we'd add a separate `MutationOperation::ProcedureOfModel`, which would be tied to a model, and whose result type would be required to be an array of the object type associated with that model. It would get an implicit argument whose type would effectively be that model's predicate type (although that no longer _needs_ to be a first-class part of the type language now). This allows us to express things like "delete where", and even most (all?) post-conditions on inserts and updates (because the check constraint would be made against the tentative state of the collection in the data source before the transaction closed, so it makes sense to express that in terms of the collection type, even if it includes things like computed fields, which wouldn't be part of the input argument type).

Yeah, looks good. I would just add that the main motivation for this approach was to tie the return type of the procedure to a model so that we can apply the select permissions for that model on the returned list of affected objects.

paf31:
As to whether predicate types should be allowed to exist on arbitrary object types (instead of merely collection types), I'm really unsure, but here are my scattered thoughts on the subject:

- Our object types are nominal, so why not assign behavior to them.
- Object types can appear in more than just input arguments - they also can appear in return types, and in the argument types of custom operators. If I hand back a predicate in a result type, I know its syntactic structure (it must be an `Expression`), but what can I do with it? Semantically, as far as the caller is concerned, is it just an abstract blob of data? Can the caller take an output predicate and feed it to any input argument of the same type, and expect any sort of behavior at all? Can the caller modify the predicate, and expect _that_ to behave in any sort particular way?
- I wanted to try to pin down the semantics of predicate types, and I fooled myself into thinking it was easier for collection types. Specifically, if you have a `Predicate<M>` in hand for some model `M`, you can at least feed it to `/query` for model `M`. You can't do this for all object types. So this gives some sort of meaning to the values of `Predicate<M>` - the denotation of a predicate is a function from query requests to row sets - but it's circular. Now you want to say, what is the meaning of `/query`? Well, it depends on the meaning of the `predicate` field in the query request, and we're back to where we started. 
- It's tempting to say that `Predicate<M>` can be given some semantics by structural induction on the predicate type, and by looking at response rows. So for example, for `and`, you say that `[[p and q]]` (`[[-]]` is the denotation) is `[[p]] ⋂ [[q]]` as row sets, `[[p or q]]` is `[[p]] ⋃ [[q]]`, etc. Now you don't have a circularity problem, and this works for object types too, but it goes wrong for `not` (SQL uses ternary logic, and `not (not p)` is not the same as `p` in general, so you can't expect too much from `not` as a unary operator), and it doesn't work for custom binary operators. `EXISTS` is probably fine because you can probably define that recursively using `/query` (not sure about column comparisons).
- We probably will want to support filtering for nested arrays at some point, which means we need predicate types for object types.

So all of this to say - my semantic concern was probably unfounded (it applies to collection types too, I think), and we may need predicates for object types in future. So we probably ought to add them now.

paf31:
Ok, this is now updated to generalize predicate types to arbitrary object types.

codedmart (approved):

PR #84, opened 2024-01-30
add alias to PathElement predicate expression

BenoitRanque:
We alias `where` expressions to `predicate`, in part because `where` is a reserved keyword in rust.

The `PathElement` does not follow this pattern.
We should make the behavior consistent. For now I've only commited the change to add the missing alias. Do note this means a change in the expected payload, so we'll need to update schemas and tests too.

Before I do that, I'd like us to consider 3 different approaches and make a descision:

1. Preserve the status quo, making things consistent. The expected json key would be `where`, and for rust the key is alliased to `predicate`
2. Change all rust keys to `r#where`. This is the standard way to get around reserved keywords, and would make it clear from the struct keys what the expected IR looks like. The IR would still expect the keys to be `where`
3. Change all keys to be `predicate`, in both the rust SDK _and_ the IR. This is the most disruptive.

I'm ambivalent about either option 2 or 3, but dislike option 1.

### DIFF
--- a/ndc-client/src/models.rs
+++ b/ndc-client/src/models.rs
@@ -496,6 +496,7 @@ pub struct PathElement {
     /// Values to be provided to any collection arguments
     pub arguments: BTreeMap<String, RelationshipArgument>,
     /// A predicate expression to apply to the target collection
+    #[serde(rename = "where")]
     pub predicate: Box<Expression>,
 }
 // ANCHOR_END: PathElement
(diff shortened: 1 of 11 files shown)


paf31:
Looks good, thanks, but can you please fix the tests, and then I'll merge this with the upcoming set of breaking changes.

paf31 (approved):

PR #85, opened 2024-01-31
RFC: explain for mutations

soupi:
[Rendered](https://github.com/hasura/ndc-spec/blob/dc131b64a3913e535e4ebb7522b5c605ca7128ed/rfcs/0004-explain-mutations.md)

This would require a change to ndc-sdk as well.


paf31 (commented):
Thank you for the PR!

I think I'd prefer to implement a separate `/explain/mutation` endpoint, possibly with its own capability, and to not break query explain unnecessarily. That way we also don't need another request type.

We'll also need a one-page RFC here, but we can add that when we finalize the plan.

soupi:
Happy to change the PR, but I think that if we want two separate endpoint, renaming `/explain` to `/explain/query` would be preferable, but I thought a single endpoint with a new request was nicer. Since we are at a point where there are several breaking changes that need to be addressed, it might be a good time to change this as well.

paf31:
Yes, let's change both now while we have the chance to make more breaking changes.

I wonder if `query/explain` is preferable or `explain/query`. Either way, the capabilities should mirror the URLs, I think.

soupi:
@paf31 I asked a few people and it seems like the majority prefer `/query/explain`. Updated the PR.

specification/src/tutorial/explain.md
```diff
@@ -1,9 +1,11 @@
 # Explain
 
-The `/explain` endpoint is not implemented in the reference implementation, simply because tje `QueryResponse` is interpreted directly in the `/query` endpoint. There is no intermediate representation (such as SQL) which could be described as a "query plan".
+The `/query/explain` and `/mutation/explain` endpoints are not implemented in the reference implementation, simply because the `QueryResponse` is interpreted directly in the `/query` endpoint.
+There is no intermediate representation (such as SQL) which could be described as a "query plan".
```
paf31, line 4:
```suggestion
```suggestion
The `/query/explain` and `/mutation/explain` endpoints are not implemented in the reference implementation, because their respective request objects are interpreted directly. There is no intermediate representation (such as SQL) which could be described as an "execution plan".
```

paf31 (dismissed):

paf31 (approved):

PR #86, opened 2024-01-31
Update Rust IR types

BenoitRanque:
Aliasing `where` keys to `predicate` in the rust types lead to confusion, this PR changes those aliases to `r#where` across the board.
This will be a breaking changes for users of the Rust IR types from this repo. This will not change the IR.

This is a follow up to #84, which _does_ change the IR (by changing one key to be `where` instead of `predicate`, for consistency)

The overall goal is to increase consistency and reduce confusion.

This PR does _not_ remove the term `predicate` as used in docs to describe what the value of the `where` key is.


BenoitRanque:
Superseeded by #87

PR #89, opened 2024-02-06
Add clarifying spec page about functions

paf31:


daniel-chambers (approved):
Looks good, thanks for the clarification. :)

I think its worth adding something to the mutations/procedures page to clarify that they are also invoked via fields (ie with __value and nested fields) in the same way as functions are.

codedmart (approved):

PR #90, opened 2024-02-07
Use NestedField to request procedure result fields

paf31:
This is the current `MutationOperationResults`:

```rust
pub struct MutationOperationResults {
    /// The number of rows affected by the mutation operation
    pub affected_rows: u32,
    /// The rows affected by the mutation operation
    pub returning: Option<Vec<IndexMap<String, RowFieldValue>>>,
}
```

Note we always return an optional list of rows.

However, this is the corresponding mutation request:

```rust
pub enum MutationOperation {
    Procedure {
        /// The name of a procedure
        name: String,
        /// Any named procedure arguments
        arguments: BTreeMap<String, serde_json::Value>,
        /// The fields to return
        fields: Option<IndexMap<String, Field>>,
    },
}
```

Fields is optional, because the result type might be a scalar.

Currently, the reference implementation packs the result into a `__value` field just like for functions, but this is a hack because `__value` was there to make functions work like collections, and there's no such constraint for procedures. The calling convention for procedures can be simpler.

It doesn't make sense to return a list of rows since we only have procedures.

## Proposal

The return type continues to be anything, so we use `NestedField` to query it (`NestedField` being the way we query single column values of arbitrary type):

```rust
pub enum MutationOperation {
    Procedure {
        /// The name of a procedure
        name: String,
        /// Any named procedure arguments
        arguments: BTreeMap<String, serde_json::Value>,
        /// The fields to return from the result, or null to return everything
        fields: Option<NestedField>,
    },
}

pub struct MutationOperationResults {
    pub result: serde_json::Value,
}
```

Basically we keep things as-is from a query expressiveness point of view, but do away with the `__value` hack.

The only thing we can't express (which would be dubious today anyway) is a relationship from a scalar return type.

### Alternative Design

Force procedures to return an object type:

```rust
pub enum MutationOperation {
    Procedure {
        /// The name of a procedure
        name: String,
        /// Any named procedure arguments
        arguments: BTreeMap<String, serde_json::Value>,
        /// The fields to return
        fields: IndexMap<String, Field>
    },
}

pub struct MutationOperationResults {
    pub result: Row,
}
```

### DIFF
--- a/ndc-reference/bin/reference/main.rs
+++ b/ndc-reference/bin/reference/main.rs
@@ -1979,19 +1979,6 @@ fn execute_upsert_article(
             Ok(old_row_fields.0)
         })?
     })
-    
-    // let returning = old_row.map(|old_row| todo!()).transpose()?;
-    // Ok(models::MutationOperationResults::Procedure {
-    //     result: serde_json::to_value(returning).map_err(|_| {
-    //         (
-    //             StatusCode::INTERNAL_SERVER_ERROR,
-    //             Json(models::ErrorResponse {
-    //                 message: "cannot encode response".into(),
-    //                 details: serde_json::Value::Null,
-    //             }),
-    //         )
-    //     })?,
-    // })
 }
 // ANCHOR_END: execute_upsert_article
 // ANCHOR: execute_delete_articles
--- a/ndc-reference/bin/reference/main.rs
+++ b/ndc-reference/bin/reference/main.rs
@@ -18,7 +18,6 @@ use indexmap::IndexMap;
 use ndc_client::models::{self, LeafCapability, RelationshipCapabilities};
 use prometheus::{Encoder, IntCounter, IntGauge, Opts, Registry, TextEncoder};
 use regex::Regex;
-use serde_json::Value;
 use tokio::sync::Mutex;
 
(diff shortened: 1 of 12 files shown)


ndc-reference/bin/reference/main.rs
```diff
@@ -1953,48 +1952,53 @@ fn execute_upsert_article(
...
+    //     result: serde_json::to_value(returning).map_err(|_| {
+    //         (
+    //             StatusCode::INTERNAL_SERVER_ERROR,
+    //             Json(models::ErrorResponse {
+    //                 message: "cannot encode response".into(),
+    //                 details: serde_json::Value::Null,
+    //             }),
+    //         )
+    //     })?,
+    // })
```
daniel-chambers, line 1994: Delete?

daniel-chambers (approved):
I like this improvement. 👍

ndc-reference/bin/reference/main.rs
paf31: Thanks!

paf31:
@codedmart Could you take a look at this one please?

daniel-chambers (dismissed):
LGTM!

PR #92, opened 2024-02-09
healthz/204 -> health/200

paf31:
Some cloud services require the health check endpoint to return `200 OK`. The `204 (No Content)` status can be (incorrectly) interpreted as failure.

This changes to require and use the `200 OK` status code to avoid unexpected issues.

### DIFF
--- a/ndc-reference/bin/reference/main.rs
+++ b/ndc-reference/bin/reference/main.rs
@@ -130,7 +130,7 @@ async fn main() {
     let app_state = Arc::new(Mutex::new(init_app_state()));
 
     let app = Router::new()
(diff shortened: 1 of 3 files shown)


codedmart (approved):

SamirTalwar (approved):

SamirTalwar:
So this requires similar changes across all NDC implementations, right?

paf31:
> So this requires similar changes across all NDC implementations, right?

No, I think just the SDK needs to change: https://github.com/hasura/ndc-hub/blob/main/rust-connector-sdk/src/connector.rs#L251

PR #95, opened 2024-02-16
Upgrade Nix and pin it to the correct version of Rust.

SamirTalwar:
I have added a `rust-toolchain.toml` file to pin the Rust version and toolchain. The Nix flake now reads this using `rust-overlay`.


paf31:
Thanks!

PR #97, opened 2024-02-17
reference: fix inconsistent latest_article schema

hgiasac:
The PR fixes the inconsistency between the schema and implementation of [latest_article function](https://github.com/hasura/ndc-spec/blob/590edfcee075de602f0e8f17ae7d575da5e82289/ndc-reference/bin/reference/main.rs#L739).  The result  is an object ([expected.json](https://github.com/hasura/ndc-spec/blob/main/ndc-reference/tests/query/get_max_article/expected.json)), but the schema is an array.


daniel-chambers (approved):
LGTM

paf31 (approved):
Thanks!

PR #98, opened 2024-02-19
fix explain API

paritosh-08:
The [previous fix](https://github.com/hasura/ndc-spec/pull/93) had a bug with incorrect appended URLs. This PR fixes that.


soupi (approved):

paf31 (approved):
Thanks!

PR #100, opened 2024-02-20
Refactor ndc-test for future development

paf31:
This PR tries to set things up so that we can work more quickly on `ndc-test` in future:

- Break up the single file structure into a module structure which follows the test structure (#96)
- Use `rand` instead of `proptest`. This is motivated by the relative complexity of writing `proptest` `Strategy` values compared to choosing values from the RNG in direct style. Since we are not making full use of `proptest` anyway (no shrinking), it doesn't pay for itself in terms of API complexity for us right now.
- Refactor snapshot code via Connector trait (#99)
- Make Reporter mutable, use nest! and test! macros (#100)


ndc-test/src/main.rs
```diff
@@ -71,15 +85,19 @@ async fn main() {
...
-                exit(1)
-            }
+            // let results = reporter.results();
+
+            // if !results.failures.is_empty() {
+            //     println!();
+            //     println!("{}", report(results));
+
+            //     exit(1)
+            // }
```
codedmart, line 100: Are these comments worth keeping around?

ndc-test/src/test_cases/query/relationships/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        super::expectations::expect_single_rows(&response)?;
+    } else {
+        eprintln!("Skipping parameterized relationship {}", foreign_key_name);
+    }
+
+    Ok(())
+}
+
+async fn select_top_n_using_foreign_key_as_array_relationship<C: Connector>(
```
codedmart, line 140: Love the name here 😆.

codedmart (dismissed):

ndc-test/src/main.rs
paf31: Oh, this needs adding back.

ndc-test/src/test_cases/query/relationships/mod.rs
paf31: They're only going to get better from here.

paf31: They're only going to get better from here.

codedmart (approved):

PR #103, opened 2024-02-26
[NS-23] ndc-test: Validate each response against its query request

paf31:
This adds some basic response validation for every test request, regardless of whether we created it automatically, or if the user wrote the test by hand.


SamirTalwar:
I tried changing ndc-postgres to verify that this is legitimate, but got into trouble because of the number of changes since the last release, so I think I'll leave it to you.

Might be worth cutting the v0.1.0 release from rc.18 and targeting these changes (and previous refactors) for v0.2.0.

paf31:
@SamirTalwar If you are running RC.18 then it should be possible to test this using `--endpoint` on the command line, even if the SDK isn't updated yet. I would like to try to know that this won't break NDC-Postgres (except for maybe uncovering a bug, who knows).

SamirTalwar:
I get errors in ndc-postgres:

```
[1] Predicates
  in Query
  in Customer
  in Simple queries
  in Predicates
Details: expected non-empty rows in RowSet

[2] Predicates
  in Query
  in Invoice
  in Simple queries
  in Predicates
Details: expected non-empty rows in RowSet
```

Unfortunately I don't see the query, so this is going to be hard to debug. Can we include that in the details?

paf31:
Yes, this was the nulls issue handled in the other PR.

I do want to log the request JSON on a failing request.

SamirTalwar:
> Yes, this was the nulls issue handled in the other PR.

Then I am happy!

SamirTalwar (approved):

codedmart (approved):

PR #106, opened 2024-02-27
Export report function on TestResults

paf31:
For `ndc-sdk`


SamirTalwar (approved):
Seems like a very benign refactor.

PR #115, opened 2024-03-06
RFC: scalar types representations/enum types

paf31:
[Rendered](https://github.com/hasura/ndc-spec/blob/eb18b82775172c819b2e071896298458c41d16e6/rfcs/0005-scalar-type-representation.md)

### DIFF
--- a/ndc-client/src/models.rs
+++ b/ndc-client/src/models.rs
@@ -104,15 +104,34 @@ pub struct SchemaResponse {
 
 // ANCHOR: ScalarType
 /// The definition of a scalar type, i.e. types that can be used as the types of columns.
+#[skip_serializing_none]
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
 #[schemars(title = "Scalar Type")]
 pub struct ScalarType {
+    ///
+    pub representation: Option<TypeRepresentation>,
     /// A map from aggregate function names to their definitions. Result type names must be defined scalar types declared in ScalarTypesCapabilities.
     pub aggregate_functions: BTreeMap<String, AggregateFunctionDefinition>,
     /// A map from comparison operator names to their definitions. Argument type names must be defined scalar types declared in ScalarTypesCapabilities.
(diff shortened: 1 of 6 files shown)


ndc-client/src/models.rs
```diff
@@ -104,16 +104,35 @@ pub struct SchemaResponse {
...
 // ANCHOR_END: ScalarType
 
+// ANCHOR: TypeHint
+///
+#[derive(
+    Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, JsonSchema,
+)]
+#[serde(tag = "type", rename_all = "snake_case")]
+#[schemars(title = "Type")]
+pub enum TypeRepresentation {
```
danieljharvey, line 127: This feels like we're inching towards connectors advertising "I know what GraphQL type this will be" without saying it.

Is the choice of types here going to restrain us when we come to putting a REST or Protobuf front end on engine, where the base scalar types might be different?

paf31: I understand the concern, that's actually why I didn't include this to begin with. But I think of this more like carving out subsets of valid response _JSON_ and nothing to do with GraphQL. I think all reasonable transport formats have corresponding ways of representing these scalar types.

paf31: > What if a scalar value is represented by a json object or array?

In that case, you wouldn't provide a type representation at all. Ideally you would give an object type to the field, but if you chose to make it a scalar (e.g. geography perhaps) then you'd just omit the representation. Another option is to add `TypeRepresentation::Opaque` and remove the `Option`.

paf31: > This would be more grounded if it were explicitly specifying a GraphQL representation type hint.

I don't think we should do this because we'd be overfitting to our one client. To me this is a clear dependency of abstractions on details (violating the D in SOLID).

danieljharvey: > I think all reasonable transport formats have corresponding ways of representing these scalar types.

This is the important thing here really isn't it, I think this change is fine.

codedmart (approved):

PR #121, opened 2024-03-19
Switch to rustls instead of native-tls for client requests.

SamirTalwar:
This removes our dependency on OpenSSL on Linux.

rustls is widely considered to be stable and reliable (perhaps even more than OpenSSL, but I am making no bets there).


danieljharvey (dismissed):

paf31 (commented):
We could pass this along as a feature to the users of these crates, something like this:

```toml
[features]
rustls = ["reqwest/rustls"]
```

What do you think?

SamirTalwar:
@paf31: Good idea. I made `native-tls` the default so this isn't a breaking change. We can switch to `rustls` on an opt-in basis.

paf31 (approved):

paf31:
Thanks!

PR #125, opened 2024-03-26
Split out an "ndc-models" crate which provides the models.

SamirTalwar:
This means that applications that do not need the client will not bring in unnecessary dependencies such as `reqwest` (and the whole HTTP stack).


paf31 (commented):
The name "models" came from the original OpenAPI generator code, and I've been meaning to change it for a long time. Maybe now is the time. Models is a fine name, and I don't really have a better one (`spec`, `types` maybe) but it's worth thinking about now.

This will also break the book generation, FYI, so those references to `models.rs` will need to be updated.

SamirTalwar:
I've fixed the book. Thanks for pointing it out.

If you can come up with a better name, I am all ears. Nothing is jumping out at me.

paf31 (approved):
Thanks!

PR #126, opened 2024-03-26
ndc-client: customize NDC response handling through a trait

rakeshkky:
Context: https://hasurahq.atlassian.net/browse/V3ENGINE-51

This PR introduces a new trait `NDCResponseHandler` to customize reading the response from `reqwest::Response`. It is useful in `v3-engine` to have a custom implementation that does the response size limitation. It fulfills the requirement for aforementioned JIRA task.


ndc-client/Cargo.toml
```diff
@@ -11,6 +11,7 @@ rustls = ["reqwest/rustls"]
 
 [dependencies]
 async-trait = "^0.1.74"
+bytes = "1.5.0"
```
SamirTalwar, line 14: Any reason not to use the latest version?

```suggestion
bytes = "1.6.0"
```

ndc-client/src/apis/default_api.rs
```diff
@@ -247,37 +218,67 @@ pub async fn schema_get(
...
+    request: reqwest::Request,
+) -> Result<T, Error> {
+    async {
+        let client = &configuration.client;
+        let mut resp = client.execute(request).await?;
+        let response_status = resp.status();
+        let response_content = match &configuration.response_size_limit {
+            Some(size_limit) => {
+                let mut size = 0;
+                let mut buf = bytes::BytesMut::new();
```
SamirTalwar, line 258: Can we use the `Content-Length` header, if present, to pre-allocate this with the correct capacity?

ndc-client/src/apis/default_api.rs
SamirTalwar, line 1: The duplication in this file is striking. Could you please follow this PR up with a change to refactor it out?

SamirTalwar (approved):

ndc-client/src/apis/default_api.rs
paf31: Is there no existing crate we can pull this from? Or even better, put it behind a trait? It doesn't seem to me like this crate needs to be every kind of client for every use case. Maybe even if engine needs a different sort of client, it could just provide its own.

paf31: I'm not sure what this adds beyond just failing fast if the `Content-Length` exceeds some limit. Isn't this what `.json()` does.

_Edit_: yes, compare with `hyper::body::to_bytes`.

rakeshkky: > Or even better, put it behind a trait?

A good idea! The `ndc-client` has to be more generic and need not do all the work. I'd prefer to have the custom logic within `v3-engine`. A trait (let's call `NdcClient`) can enable to do this.

> I'm not sure what this adds beyond just failing fast if the Content-Length exceeds some limit

:100: True

rakeshkky: I found a [`Limited`](https://docs.rs/http-body-util/latest/http_body_util/struct.Limited.html) http body impl in `http_body_util` crate. We could use this to stream the response with a limit on the size. I can't just use it since the `reqwest` restricts accessing inner `res: hyper::Response` of `reqwest::Response` (or) the `reqwest::Response` does not implement `hyper::Body`. 

I feel, the `reqwest` don't want us to bother much about internal (low-level) client details. Now we have either `chunk` or `bytes_stream` methods to incrementally consume the response.

ndc-client/src/apis/default_api.rs
rakeshkky: Created a task to follow-up: https://hasurahq.atlassian.net/browse/APG-53

ndc-client/src/apis/mod.rs
```diff
@@ -92,6 +95,30 @@ pub fn urlencode<T: AsRef<str>>(s: T) -> String {
     ::url::form_urlencoded::byte_serialize(s.as_ref().as_bytes()).collect()
 }
 
+/// Trait to handle `reqwest::Response`, a response returned by an NDC, by parsing it as a JSON value.
+///
+/// Useful for implementing custom NDC response handling, such as limiting the response by size etc.
+#[async_trait::async_trait]
+pub trait NDCResponseHandler: Sync {
```
SamirTalwar, line 102: Drop the `NDC` prefix; everything in this repository is NDC.

SamirTalwar (dismissed):

paf31:
Let's hold off on this for now. I've discussed this with a few people and I'm going to make an RFC to move the client out of this repo. Engine can maintain its own client code, and `ndc-test` can use the existing code here (minus tracing).

paf31:
Closing in favor of #130.

PR #128, opened 2024-03-27
Upgrade to Rust v1.77, and fix the ndc-reference Docker image.

SamirTalwar:
This makes a few changes to get the ndc-reference Docker image working and up to speed:

- the correct Debian version is used to avoid glibc errors at runtime
- we use `cargo-chef` to avoid rebuilding dependencies
- we now conform to the connector spec when selecting the port
- we bind the host to any IPv4 _or_ IPv6 address, not just IPv4
- the sample JSON is bundled into the binary rather than being read at runtime, to avoid lookup failures
- a shutdown handler is installed to catch Ctrl+C and friends

I've also upgraded to Rust 1.77.


ndc-reference/bin/reference/main.rs
```diff
@@ -143,11 +148,43 @@ async fn main() {
         ))
         .with_state(app_state);
 
-    // run it with hyper on localhost:8100
-    axum::Server::bind(&"0.0.0.0:8100".parse().unwrap())
-        .serve(app.into_make_service())
-        .await
-        .unwrap();
+    // start the server on localhost:<PORT>
+    let port = env::var("HASURA_CONNECTOR_PORT")
```
paf31, line 152: Are these changes necessary? There's no requirement for the reference agent to implement the deployment spec.

paf31: I know it seems pedantic, but I'd really like the reference agent to be as minimal as possible. Also, I'd like it to be an example of what is _not_ required in order to meet the spec.

SamirTalwar: We use it for testing in the engine, so it's very useful to be able to run it on a given port. Right now we do this inside Docker but we'd like to make it possible to run tests without relying on Docker for our own code.

ndc-reference/bin/reference/main.rs
```diff
@@ -144,13 +149,48 @@ async fn main() {
...
-    // run it with hyper on localhost:8100
-    axum::Server::bind(&"0.0.0.0:8100".parse().unwrap())
-        .serve(app.into_make_service())
-        .await
-        .unwrap();
+    // Start the server on `localhost:<PORT>`.
+    // This says it's binding to an IPv6 address, but will actually listen to
+    // any IPv4 or IPv6 address.
+    let host = net::IpAddr::V6(net::Ipv6Addr::UNSPECIFIED);
+    let port = env::var("PORT")
```
paf31, line 156: I changed this to `PORT` to avoid any confusion that we might be implementing the deployment spec, and factored out a `shutdown_handler()` function, since `main()` gets shown in full in the book. Otherwise, looks good to me, merging now, thanks!

paf31 (dismissed):

PR #129, opened 2024-04-02
RFC: more type representations

paf31:
[Rendered](https://github.com/hasura/ndc-spec/blob/20b3b2d9e26d5b0a950a679020cb7d9674aa3644/rfcs/0007-additional-type-representations.md)


rfcs/0007-additional-type-representations.md
```diff
@@ -0,0 +1,48 @@
...
+|-|-|-|
+| Int8 | A 8-bit signed integer with a minimum value of -2^7 and a maximum value of 2^7 - 1 | Number |
+| Int16 | A 16-bit signed integer with a minimum value of -2^15 and a maximum value of 2^7 - 1 |Number |
+| Int32 | A 32-bit signed integer with a minimum value of -2^31 and a maximum value of 2^7 - 1 |Number |
+| Int64 | A 64-bit signed integer with a minimum value of -2^63 and a maximum value of 2^7 - 1 |String |
+| Float32 | An IEEE-754 single-precision floating-point number | Number |
+| Float64 | An IEEE-754 double-precision floating-point number | String |
+| Decimal | | String |
+| Char | A unicode scalar value |
+| Uuid | | String |
```
hgiasac, line 24: IMO,  I prefer `UUID` instead
`Int` already represents `Int32`.  We don't need to define another type.  `Number` is also duplicated with `Float64`. Should we deprecate it?
FYI this is also a good reference https://the-guild.dev/graphql/scalars/docs/scalars

rfcs/0007-additional-type-representations.md
```diff
@@ -0,0 +1,48 @@
...
+| Date | | String |
+| Timestamp | | String |
+| TimestampTZ | | String (ISO 8601) |
+| Bytes | | Base64-encoded String |
+| Geography | Geometry | String (GeoJSON) |
```
hgiasac, line 29: btw, should we consider adding a JSON type for dynamic json objects? It's better to avoid decoding twice (string -> byte[] -> Geography object)

rfcs/0007-additional-type-representations.md
```diff
@@ -0,0 +1,48 @@
...
+## Proposal
+
+Add the following type representations to `ndc_models::TypeRepresentation`:
+
+|Name|Description|JSON representation|
+|-|-|-|
+| Int8 | A 8-bit signed integer with a minimum value of -2^7 and a maximum value of 2^7 - 1 | Number |
+| Int16 | A 16-bit signed integer with a minimum value of -2^15 and a maximum value of 2^7 - 1 |Number |
+| Int32 | A 32-bit signed integer with a minimum value of -2^31 and a maximum value of 2^7 - 1 |Number |
+| Int64 | A 64-bit signed integer with a minimum value of -2^63 and a maximum value of 2^7 - 1 |String |
```
soupi, line 19: I think the maximum needs updating here.

```suggestion
| Int16 | A 16-bit signed integer with a minimum value of -2^15 and a maximum value of 2^15 - 1 |Number |
| Int32 | A 32-bit signed integer with a minimum value of -2^31 and a maximum value of 2^31 - 1 |Number |
| Int64 | A 64-bit signed integer with a minimum value of -2^63 and a maximum value of 2^63 - 1 |String |
```

rfcs/0007-additional-type-representations.md
```diff
@@ -0,0 +1,48 @@
...
+- We want to be able to generate some standard scalar types in Hasura for types which are used across different data sources, e.g. UUIDs. This needs support in NDC.
+
+## Proposal
+
+Add the following type representations to `ndc_models::TypeRepresentation`:
+
+|Name|Description|JSON representation|
+|-|-|-|
+| Int8 | A 8-bit signed integer with a minimum value of -2^7 and a maximum value of 2^7 - 1 | Number |
+| Int16 | A 16-bit signed integer with a minimum value of -2^15 and a maximum value of 2^7 - 1 |Number |
```
plcplc, line 17: > ... and a maximum value of 2^7 -1

This looks like a copy-paste error from Int8

rfcs/0007-additional-type-representations.md
```diff
@@ -0,0 +1,48 @@
+# Additional Type Representations
+
+## Purpose
+
+The current list of scalar type representations is inadequate:
+
+- JSON integer and number types are not what we want for most applications, since they are represented by an unusual 53-bit number type. Since scalars can be both inputs and outputs, and are therefore invariant, we cannot safely convert to and from larger and smaller integer and floating point types.
+- We want to be able to generate some standard scalar types in Hasura for types which are used across different data sources, e.g. UUIDs. This needs support in NDC.
```
plcplc, line 8: Why does this influence type representations?

As far as I understand the purpose of type representations, they essentially only serve to fill a gap in query/result encoding that is underspecfied without them, i.e. they help answering the question of: "Given some selection field of some scalar type, what kind of literal can I expect in the query result for this field?" and vice versa for input fields. Is this view of their use is too narrow?

If the purpose is to allow e.g. remote relationships between different hosted graphql service over, say, a common `UUID` type, isn't that solved by simply picking the name `UUID` for the scalar type and hoping/asking/ensuring it be represented as a string by the ndc?

rfcs/0007-additional-type-representations.md
```diff
@@ -0,0 +1,48 @@
...
+Add the following type representations to `ndc_models::TypeRepresentation`:
+
+|Name|Description|JSON representation|
+|-|-|-|
+| Int8 | A 8-bit signed integer with a minimum value of -2^7 and a maximum value of 2^7 - 1 | Number |
+| Int16 | A 16-bit signed integer with a minimum value of -2^15 and a maximum value of 2^7 - 1 |Number |
+| Int32 | A 32-bit signed integer with a minimum value of -2^31 and a maximum value of 2^7 - 1 |Number |
+| Int64 | A 64-bit signed integer with a minimum value of -2^63 and a maximum value of 2^7 - 1 |String |
+| Float32 | An IEEE-754 single-precision floating-point number | Number |
+| Float64 | An IEEE-754 double-precision floating-point number | String |
```
plcplc, line 21: I feel like having both a `Name` and a `JSON representation` column conflates concerns.
I'm specifically not too hot on these `| Int64 | .. | String |` and `| Float64 | .. | String |` rows, as they rule out actually having `Int64` and `Float64` be represented by numbers, which is a legitimate thing to want to do.

rfcs/0007-additional-type-representations.md
paf31: We can always just omit a representation for types which are represented by some opaque JSON.

rfcs/0007-additional-type-representations.md
paf31: `Int` is a JSON 53-bit integer, and since types can be used as inputs and outputs, the concern was that these are semantically imprecise.

A connector author could choose to use `Int` for a 32-bit integer, but would have to throw an error if that type were reused as an input (e.g. in a predicate) and an out-of-range value were provided.

I actually think that's a reasonable trade-off for the simplicity, but we can discuss it here.

rfcs/0007-additional-type-representations.md
```diff
@@ -0,0 +1,48 @@
+# Additional Type Representations
+
+## Purpose
+
+The current list of scalar type representations is inadequate:
+
+- JSON integer and number types are not what we want for most applications, since they are represented by an unusual 53-bit number type. Since scalars can be both inputs and outputs, and are therefore invariant, we cannot safely convert to and from larger and smaller integer and floating point types.
```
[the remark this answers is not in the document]
paf31: That's encouraging, I think we could test that with a few clients and see what sort of support it has.

rfcs/0007-additional-type-representations.md
paf31: > Why does this influence type representations?

Good question. Representations were initially intended as a way to communicate hints for code generation, e.g. GraphQL schemas. But now we are planning to generate a fixed set of scalars on the engine side, this is the obvious place to indicate those as well, where it makes sense to do so.

That is arguably violating the "D" in SOLID, since the server spec is now dependent on client details. The alternative is an open world of scalars, in which engine publishes the scalars it supports, and there is some coupling between connectors using those scalars and engine's definition of them. Which arguably violates "I" in SOLID, since servers are forced to implement (parts of) interfaces they may not use.

rfcs/0007-additional-type-representations.md
paf31: A scalar can be represented as anything including something not in this list. That is why `representation` is optional.

rfcs/0007-additional-type-representations.md
paf31: After a discussion, I have updated the spec to remove `Int` and `Float`.

rfcs/0007-additional-type-representations.md
paf31: I've added `JSON` as a representation, and we will make representation required.

rfcs/0007-additional-type-representations.md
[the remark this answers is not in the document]
paf31: @0x777 had some concerns that actual support for arbitrary integers in clients was very limited. Also, we would like not to be too tied to JSON as a transport mechanism anyway. I've tried to call this out a bit more in the latest revision.

rfcs/0007-additional-type-representations.md
paf31: Update: representation will not be optional any more.

> as they rule out actually having Int64 and Float64 be represented by numbers, which is a legitimate thing to want to do

Legitimate but not possible in JSON with good client support, as I understand it. Later, we can add a GRPC transport, and appropriate representations for each of these types using the GRPC wire protocol, including int64s as actual integers.

rfcs/0007-additional-type-representations.md
```diff
@@ -0,0 +1,58 @@
...
+Add the following type representations to `ndc_models::TypeRepresentation`:
+
+|Name|Description|JSON representation|
+|-|-|-|
+| Int8 | A 8-bit signed integer with a minimum value of -2^7 and a maximum value of 2^7 - 1 | Number |
+| Int16 | A 16-bit signed integer with a minimum value of -2^15 and a maximum value of 2^15 - 1 | Number |
+| Int32 | A 32-bit signed integer with a minimum value of -2^31 and a maximum value of 2^31 - 1 | Number |
+| Int64 | A 64-bit signed integer with a minimum value of -2^63 and a maximum value of 2^63 - 1 | String |
+| Float32 | An IEEE-754 single-precision floating-point number | Number |
+| Float64 | An IEEE-754 double-precision floating-point number | String |
```
abhinav-hasura, line 24: Shouldn't Float64 be Number instead of String?
In v2, float64 postgres columns are returned as JSON numbers.
Javascript number is also a Float64 so I think all libraries should be able to interpret number as a Float64.

abhinav-hasura (dismissed):

ndc-models/src/lib.rs
```diff
@@ -131,9 +136,39 @@ pub enum TypeRepresentation {
...
+    /// A 16-bit signed integer with a minimum value of -2^15 and a maximum value of 2^15 - 1
+    Int16,
+    /// A 32-bit signed integer with a minimum value of -2^31 and a maximum value of 2^31 - 1
+    Int32,
+    /// A 64-bit signed integer with a minimum value of -2^63 and a maximum value of 2^63 - 1
+    Int64,
+    /// An IEEE-754 single-precision floating-point number
+    Float32,
+    /// Arbitrary-precision decimal string
+    Decimal,
```
abhinav-hasura, line 153: nit: You mentioned calling this BigDecimal would be clearer right?

abhinav-hasura (approved):

ndc-models/src/lib.rs
```diff
@@ -131,9 +134,39 @@ pub enum TypeRepresentation {
...
+    /// A 8-bit signed integer with a minimum value of -2^7 and a maximum value of 2^7 - 1
+    Int8,
+    /// A 16-bit signed integer with a minimum value of -2^15 and a maximum value of 2^15 - 1
+    Int16,
+    /// A 32-bit signed integer with a minimum value of -2^31 and a maximum value of 2^31 - 1
+    Int32,
+    /// A 64-bit signed integer with a minimum value of -2^63 and a maximum value of 2^63 - 1
+    Int64,
+    /// An IEEE-754 single-precision floating-point number
+    Float32,
```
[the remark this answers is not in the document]
paf31: Thanks!

ndc-models/src/lib.rs
paf31: yes I'll modify this.

ndc-models/src/lib.rs
```diff
@@ -109,7 +111,8 @@ pub struct SchemaResponse {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
 #[schemars(title = "Scalar Type")]
 pub struct ScalarType {
-    /// An optional description of valid values for this scalar type
+    /// A description of valid values for this scalar type.
+    /// Defaults to `TypeRepresentation::Json` if omitted
```
abhinav-hasura, line 115:
```suggestion
    /// Defaults to `TypeRepresentation::JSON` if omitted
```

abhinav-hasura (approved):

hgiasac:
I remember about unsigned integers. Just wonder if we support `uint` representations

paf31:
> uint representations

Not right now, but we can consider adding them if there's a use case you have in mind?

hgiasac:
Not really. `Int32` and `Int64` types can cover most of the use cases. However, connector authors should be careful when using `uint` types. For example, the value range of `int32` is `-2^31 to 2^31 - 1`, but the value range of `uint32` is `0 to 2^63 - 1`. Therefore we should use `int64` representation for `uint32` to avoid data lost if the value > 2^31.

PR #133, opened 2024-04-04
add ndc-models to root Cargo.toml

hgiasac:
`ndc-sdk-typescript` generates jsonschema from `ndc-models`. This package should be exposed in the root Cargo.toml


paf31 (approved):
Thanks!

PR #135, opened 2024-04-04
Add back features for native-tls vs rustls

paf31:
@SamirTalwar I realized this was lost in the recent removal of `ndc-client`


SamirTalwar:
@paf31: Yeah, I didn't flag it because it seemed unnecessary to worry about this just for ndc-test. I cared about it for ndc-client.

I am happy for this to be merged but I don't particularly mind if it's not.

PR #136, opened 2024-04-10
Added a biginteger type representation

daniel-chambers:
Adds a `biginteger` type representation, which is like `bigdecimal` but only contains integers. 
This is useful to represent JavaScript's [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#bigint_type) type.


paf31 (approved):
Thanks, just need to update the test files (I'll do it)

daniel-chambers:
Ah bugger, I should have checked that. Thanks for updating the tests @paf31. Looks like you need to re-approve.

paf31 (approved):

PR #138, opened 2024-05-03
Filtering and ordering by values of nested fields

dmoverton:
This PR replaces https://github.com/hasura/ndc-spec/pull/27.  Please refer to [RFC 0009](https://github.com/hasura/ndc-spec/blob/dmoverton/nested-filter-order-by/rfcs/0009-filter-and-order-by-nested-fields.md), which forms part of this PR, for details.

### DIFF
--- a/ndc-models/src/lib.rs
+++ b/ndc-models/src/lib.rs
@@ -61,7 +61,7 @@ pub struct QueryCapabilities {
     /// Does the connector support explaining queries
     pub explain: Option<LeafCapability>,
     /// Does the connector support nested fields
-    pub nested_fields: Option<NestedFieldCapabilities>,
+    pub nested_fields: NestedFieldCapabilities,
 }
 // ANCHOR_END: QueryCapabilities
 
--- a/ndc-models/src/lib.rs
+++ b/ndc-models/src/lib.rs
@@ -503,7 +503,7 @@ pub enum OrderByTarget {
         /// The name of the column
         name: String,
         /// Path to a nested field within an object column
-        column_path: Option<Vec<String>>,
+        field_path: Option<Vec<String>>,
         /// Any relationships to traverse to reach this column
         path: Vec<PathElement>,
     },
@@ -585,7 +585,7 @@ pub enum ComparisonTarget {
         /// The name of the column
         name: String,
         /// Path to a nested field within an object column
-        column_path: Option<Vec<String>>,
+        field_path: Option<Vec<String>>,
         /// Any relationships to traverse to reach this column
         path: Vec<PathElement>,
     },
@@ -593,7 +593,7 @@ pub enum ComparisonTarget {
         /// The name of the column
         name: String,
         /// Path to a nested field within an object column
-        column_path: Option<Vec<String>>,
+        field_path: Option<Vec<String>>,
     },
 }
 // ANCHOR_END: ComparisonTarget
--- a/ndc-models/src/lib.rs
+++ b/ndc-models/src/lib.rs
@@ -501,7 +501,7 @@ pub struct OrderByElement {
 pub enum OrderByTarget {
     Column {
         /// The name of the column
-        name: String,
+        name: ColumnSelector,
         /// Any relationships to traverse to reach this column
         path: Vec<PathElement>,
     },
@@ -573,6 +573,16 @@ pub enum UnaryComparisonOperator {
 }
 // ANCHOR_END: UnaryComparisonOperator
 
+// ANCHOR: ColumnSelector
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(untagged)]
+pub enum ColumnSelector {
+    Path(Vec<String>),
+    Column(String),
+}
+// ANCHOR_END: ColumnSelector
+
+
 // ANCHOR: ComparisonTarget
 #[skip_serializing_none]
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
@@ -581,13 +591,13 @@ pub enum UnaryComparisonOperator {
 pub enum ComparisonTarget {
     Column {
         /// The name of the column
-        name: String,
+        name: ColumnSelector,
         /// Any relationships to traverse to reach this column
         path: Vec<PathElement>,
     },
     RootCollectionColumn {
         /// The name of the column
-        name: String,
+        name: ColumnSelector,
     },
 }
 // ANCHOR_END: ComparisonTarget
--- a/ndc-models/src/lib.rs
+++ b/ndc-models/src/lib.rs
@@ -60,8 +60,22 @@ pub struct QueryCapabilities {
     pub variables: Option<LeafCapability>,
     /// Does the connector support explaining queries
     pub explain: Option<LeafCapability>,
+    /// Does the connector support nested fields
+    pub nested_field_capabilities: Option<NestedFieldCapabilities>,
 }
 // ANCHOR_END: QueryCapabilities
+
+// ANCHOR: NestedFieldCapabilities
+#[skip_serializing_none]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Nested Field Capabilities")]
+pub struct NestedFieldCapabilities {
+    /// Does the connector support filtering by values of nested fields
+    pub filter_by_nested_fields: Option<LeafCapability>,
+    /// Does the connector support ordering by values of nested fields
+    pub order_by_nested_fields: Option<LeafCapability>,
+}
+// ANCHOR_END: NestedFieldCapabilities
 
 // ANCHOR: MutationCapabilities
 #[skip_serializing_none]
--- a/ndc-models/src/lib.rs
+++ b/ndc-models/src/lib.rs
@@ -501,7 +501,9 @@ pub struct OrderByElement {
 pub enum OrderByTarget {
     Column {
         /// The name of the column
-        name: ColumnSelector,
+        name: String,
+        /// Path to a nested field within an object column
+        column_path: Option<Vec<String>>,
         /// Any relationships to traverse to reach this column
         path: Vec<PathElement>,
     },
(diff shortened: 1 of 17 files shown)


specification/src/specification/queries/filtering.md
```diff
@@ -69,6 +69,9 @@ This example uses a custom `like` operator:
 
 Comparison operators compare columns to values. The column on the left hand side of any operator is described by a [`ComparisonTarget`](../../reference/types.md#comparisontarget), and the various cases will be explained next.
 
+The `name` property of `ComparisonTarget` is a [`ColumnSelector`](../../reference/types.md#columnselector), which can be either a string or an array of strings.  If it is a string or a singleton array then it refers to a scalar-valued top-level column. `ColumnSelector`s with multi-valued arrays are only relevant for databases that support nested objects, e.g. MongoDB. If `name` is a multi-value array then the first element refers to an object-valued top-level column and the remaining elements specify a path to a scalar-valued field within a nested object within that column.
```
paf31, line 72: Everything else uses an implicit `exists` when array traversals are involved. We don't _need_ to support that use case now, but I just wanted to note it here.

ndc-reference/bin/reference/main.rs
```diff
@@ -1236,14 +1237,46 @@ fn eval_order_by_single_column_aggregate(
...
+
 // ANCHOR: eval_order_by_column
 fn eval_order_by_column(
     collection_relationships: &BTreeMap<String, models::Relationship>,
     variables: &BTreeMap<String, serde_json::Value>,
     state: &AppState,
     item: &BTreeMap<String, serde_json::Value>,
     path: Vec<models::PathElement>,
-    name: String,
+    name: &ColumnSelector,
```
paf31, line 1279: Can we rename this to `selector`?

ndc-reference/bin/reference/main.rs
```diff
@@ -1256,7 +1289,7 @@ fn eval_order_by_column(
         ));
     }
     match rows.first() {
-        Some(row) => eval_column(row, name.as_str()),
+        Some(row) => eval_select_from_row(row, name),
```
paf31, line 1292: Following the naming elsewhere, let's please call this `eval_column_selector`.

ndc-reference/bin/reference/main.rs
```diff
@@ -1236,14 +1237,46 @@ fn eval_order_by_single_column_aggregate(
...
+        [] => Some(value),
+        [head, tail @ ..] => value.get(head).and_then(|v| get_json_path(v, tail)),
+    }
+}
+
+fn eval_select_from_row(
+    row: &Row,
+    column_selector: &ColumnSelector,
+) -> Result<serde_json::Value> {
+    match column_selector {
```
paf31, line 1255: This looks like a reduce operation?

ndc-models/src/lib.rs
```diff
@@ -573,6 +573,16 @@ pub enum UnaryComparisonOperator {
 }
 // ANCHOR_END: UnaryComparisonOperator
 
+// ANCHOR: ColumnSelector
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(untagged)]
```
paf31, line 578: There's a couple of things I'd like to change here:

- We don't use `untagged` generally, so I'd like to avoid it.
- The `Path` String vector consists of two different types of string (names of fields, and names of JSON fields), so I'd like to change that.

So here is what I would suggest:

- Everywhere `ColumnSelector` is used, add an adjacent field of type `Option<Vec<String>>`, meaning the list of JSON field names to traverse.
- Remove `ColumnSelector`.

So the idea is we always have a _column_ name, and we _might_ have a list of JSON field names to traverse.

The capability would enable that optional field.

paf31 (changes requested):
Thank you for the detailed PR and RFC :)

I have a few suggestions, and I'd also like to please add a new pair of capabilities to track this.

Can you also please add some reference implementation snapshot tests?

dmoverton:
> Thank you for the detailed PR and RFC :)
> 
> I have a few suggestions, and I'd also like to please add a new pair of capabilities to track this.
> 
> Can you also please add some reference implementation snapshot tests?

@paf31 I have addressed your review comments by removing `ColumnSelector` and replacing it with new property `field_path` with type `Option<Vec<String>>`. Are you happy with that name?

I have not yet added the new capabilities or snapshot test. I will address these tommorrow.

ndc-reference/bin/reference/main.rs
```diff
@@ -1236,6 +1246,30 @@ fn eval_order_by_single_column_aggregate(
...
+
+// ANCHOR: eval_column_field_path
+fn eval_column_field_path(
+    row: &Row,
+    column_name: &String,
+    field_path: &Option<Vec<String>>,
+) -> Result<serde_json::Value> {
+    match field_path {
+        None => eval_column(row, column_name),
+        Some(path) => row
```
paf31, line 1258: Let's reuse `eval_column` here for the better error message, and `?` over `and_then`, please, to be consistent.

dmoverton: Done

dmoverton:
@paf31 I have added the capabilities and tests. Could you please re-review?

specification/src/specification/queries/sorting.md
```diff
@@ -16,6 +16,11 @@ To compute the ordering from the `order_by` field, data connectors should implem
 
 ### Type `column`
 
+The property `element.target.name` refers to a column name.
```
paf31, line 19: I think this still needs to be moved to its own subsection of one or the other page, and linked from the other, please.

ndc-reference/tests/capabilities/expected.json
```diff
@@ -3,7 +3,11 @@
   "capabilities": {
     "query": {
       "aggregates": {},
-      "variables": {}
+      "variables": {},
+      "nested_field_capabilities": {
```
paf31, line 7: Let's call this `nested_fields` please, we don't repeat `_capabilities` elsewhere. Same for below, I think, I think we can just call the sub-capabilities `filter_by` and `order_by` since they are scoped by inclusion.

Also, I think we should either make `nested_field_capabilities` required, since it has no meaning on its own, or lift `query.nested_fields.filter_by` to `query.filter_by_nested_fields` etc.

We also need to document these new capabilities in the spec, on the two new pages, and on the capabilities page, please.

paf31 (commented):
Basically looks great, I have a couple of small-ish notes, but I think the main thing to do now is to test this in engine.

codedmart:
> Basically looks great, I have a couple of small-ish notes, but I think the main thing to do now is to test this in engine.

We need to test this in engine before merging? That would take implementing it in the engine first right? I know at least Elastic and Mongo will use this, and would be good if the spec was in for those connectors to start working against.

paf31:
> We need to test this in engine before merging? That would take implementing it in the engine first right? I know at least Elastic and Mongo will use this, and would be good if the spec was in for those connectors to start working against.

I do think it would be good to make a POC engine PR, even just an RFC, to make sure this is usable with a reasonable GraphQL schema. I'd like to avoid adding more unused or partially-used features which add to the complexity of the spec for all authors, but which don't necessarily get used. Connector authors can always build against a branch, right?

specification/src/specification/queries/sorting.md
dmoverton: I have done this. I'm not sure it really improves the readability of the spec though.

codedmart:
> > We need to test this in engine before merging? That would take implementing it in the engine first right? I know at least Elastic and Mongo will use this, and would be good if the spec was in for those connectors to start working against.
> 
> I do think it would be good to make a POC engine PR, even just an RFC, to make sure this is usable with a reasonable GraphQL schema. I'd like to avoid adding more unused or partially-used features which add to the complexity of the spec for all authors, but which don't necessarily get used. Connector authors can always build against a branch, right?

OK fair, @dmoverton let's chat about next steps here.

paf31:
To be clear, I'm not saying build the entire engine feature, just to make sure we can import these new types, and generate the IR and something that looks like the GraphQL schema we want to see. Having an RFC for the GraphQL schema seems pretty essential though.

specification/src/specification/queries/filtering.md
```diff
@@ -85,6 +85,12 @@ If the `ComparisonTarget` has type `root_collection_column`, then the `name` pro
 
 The root collection is defined as the collection in scope at the nearest enclosing [`Query`](../../reference/types.md#query), and the column should be chosen from the _row_ in that collection which was in scope when that `Query` was being evaluated.
 
+#### `field_path`
```
paf31, line 88:
```suggestion
#### Referencing nested fields within columns
```

specification/src/specification/queries/sorting.md
```diff
@@ -16,6 +16,9 @@ To compute the ordering from the `order_by` field, data connectors should implem
...
+If the connector supports capability `query.nested_fields.order_by` then the target may also have a [`field_path`](./filtering.md#field_path) property.
```
paf31, line 20: Sorry if we change the title, this will need to be updated too.

specification/src/specification/queries/sorting.md
paf31: I do think it's useful to have a fragment that we can link to.

ndc-models/src/lib.rs
soupi: Is there a meaningful difference between `Option<Vec<String>>` and `Vec<String>`? Not sure why we prefer the former over the latter.

Also, I would suggest having `Vec<String>` be a newtype `struct Path(Vec<String>)`.

paf31: The idea was that we want to implement this as a non-breaking change, but we might be able to do that with `skip_serializing_if`.

ndc-models/src/lib.rs
```diff
@@ -502,6 +516,8 @@ pub enum OrderByTarget {
     Column {
         /// The name of the column
         name: String,
+        /// Path to a nested field within an object column
+        field_path: Option<Vec<String>>,
```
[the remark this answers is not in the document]
paf31: @dmoverton Let's try using `skip_serializing_if` to make this null/empty (empty not serialized) or not-null and non-empty. We just need to make sure the JSON schema is generated correctly.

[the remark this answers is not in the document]
paf31: Actually it looks like this affects the JSON schema, so let's keep it as-is.

paf31 (approved):

PR #139, opened 2024-05-06
Field Arguments RFC/Spec

sordina:
This PR updates the NDC-Spec to allow arguments to fields.

See the [rendered RFC](https://github.com/hasura/ndc-spec/blob/lyndon/field-arguments/rfcs/0009-field-arguments.md) here.

The motivation for this change is to generalise the API surface to allow more expressive queries - This should be very similar to collection arguments in practice but can apply to any field, not just top-level collections.

Some examples of why this might be useful:

* JSON Operations: For a JSON field in a PG table, several JSON functions could be exposed as fields but they require arguments to be useful. Such as [#>](https://www.postgresql.org/docs/9.3/functions-json.html)
* Vector Operations for LLMs etc.
* Advanced Geo Operations
* GraphQL federation: Forwarding Hasura GQL schemas over NDC will require this change if we want to expose root fields as commands instead of collections.
* etc.

## Changes

The following changes are present in this PR:

* An RFC has been added detailing this change
* Documentation in the Spec has been updated detailing this functionality
* Schema datatypes now support defining field arguments in schemas
    * Main change: `ObjectField` in `ndc-models/src/lib.rs`
    * Copying pattern from `CollectionInfo` of `BTreeMap<String, ArgumentInfo>`
    * This field is optional for backwards compatibility
* Query datatypes now support issuing queries with field arguments
    * `Field` extended with `arguments: BTreeMap<String, Argument>`
    * As per `QueryRequest` for collections
* The reference connector now uses field arguments for testing purposes
    * Added limit to nested arrays
    * Added limit argument to the `institutions`.`staff` array field.
* The golden test suite makes assertions about the field arguments used by the reference agent
* NDC Spec now omits fields with required arguments

## Follow Up Work

* Engine changes

## Notes

* Schema and Query should probably be split into separate files to make it clear which phase the code is operating on but this should be done in a follow-up PR to make this change as clear as possible
* New schema and query fields have been made optional via serde parser directives in order to allow them to be omitted for backwards compatibility reasons.


paf31 (commented):
Looks great so far, thanks. Couple of notes:

- I think we could come up with a better example than truncating the title: what about adding a `where` or `limit` argument to the nested arrays within `institution`? That would mirror a real use case.
- You'll also need to update the test cases for the reference agent to exercise these new code branches.
- `ndc-test` will have to be changed to omit any fields which take arguments, since we don't generate argument values in general.

rfcs/0010-field-arguments.md
```diff
@@ -0,0 +1,57 @@
...
+
+This mirrors the existing implementation for collection arguments with the
+addition of an `Option` wrapper in order to make the implementation backwards
+compatible. RFC for this aspect - Should these be non-optional?
+
+## Implications
+
+NDC schema and query invocation:
+
+* When the schema indicates that a field has arguments then they may be provided in a query.
```
paf31, line 49: Let's please take the same approach as other types of arguments and say that all arguments are required whether nullable or not.

sordina: Couldn't this bloat the payload significantly?

sordina: Especially as we start to leverage common field arguments such as limit, etc. A query which has no field constraints could then require every argument to be specified as null explicitly.

sordina: I guess we could consider this an optimisation concern and address later along with other similar issues.

paf31: Yes it's probably not necessary, but let's be consistent for now, and we can relax this later for all types of nullable input arguments at once.

ndc-models/src/lib.rs
```diff
@@ -205,6 +205,9 @@ pub struct ObjectField {
     /// The type of this field
     #[serde(rename = "type")]
     pub r#type: Type,
+    /// The arguments available to the field - Matches implementation from CollectionInfo
+    #[serde(skip_serializing_if = "Option::is_none")]
```
paf31, line 209: You already have `skip_serializing_none` on the `struct`.

ndc-reference/bin/reference/main.rs
```diff
@@ -1833,10 +1870,37 @@ fn eval_nested_field(
...
                 )
             })?;
-            let result_array = array
+            let mut limit = array.len();
+            let limit_error = (StatusCode::INTERNAL_SERVER_ERROR,
+                                Json(models::ErrorResponse {
+                                    message: "Cannot decode `limit` argument".into(),
+                                    details: serde_json::Value::Null, }));
+            if let Some(map) = arguments {
+                println!("Unevaluated arguments: {:?}", arguments);
```
paf31, line 1879: Debug code?

ndc-reference/bin/reference/main.rs
```diff
@@ -1833,10 +1870,37 @@ fn eval_nested_field(
...
+
+                if let Some(unevaluated_argument) = map.get("limit") {
+                    // limit = value;
+                    let evaluated_argument = eval_argument(variables, unevaluated_argument)?;
+                    println!("Evaluated argument: {:?}", evaluated_argument);
+                    match evaluated_argument {
+                        Value::Number(n) => {
```
paf31, line 1886: You can use `value.as_i64()` here, there are several examples throughout this file. Can you also please read `limit` once, like we do for other input examples.

E.g.

```rust
let id: i32 = row
            .get("id")
            .ok_or("'id' field not found in json file")?
            .as_i64()
            .ok_or("'id' field was not an integer in json file")?
            .try_into()?;
```

paf31 (changes requested):
We need to make the `ndc-test` change to omit fields with arguments, before this can merged, or else we'll start generating invalid requests.

The specification should also be updated please.

ndc-reference/bin/reference/main.rs
sordina: Sure thing.

sordina: Actually I cast as usize since this is used by arrays to for bounds, so the marshalling here should do an extra error check for free. What do you think?

sordina: `let mut limit = array.len();`

ndc-reference/bin/reference/main.rs
sordina: yep

ndc-models/src/lib.rs
sordina: great!

sordina:
@paf31 there's a recursive case when if the only field in a query is omitted due to arguments, then the parent should also be omitted, etc.

sordina:
@paf31 I think I've got everything updated:

* Fields are omitted from tests if they have any required arguments
* Arguments are now required to be explicitly null
* Debugging removed

I'm not sure how to use the ok_or pattern you recommended since we don't actually want to throw errors in the case of Nones in the reference agent unless I'm missing something. Let me know if you've got a sec to go over this.

rfcs/0010-field-arguments.md
sordina: 👍

paf31 (commented):
@sordina I'm probably not going to have time this evening to go through this together, so I pushed a commit to the top of the branch showing how I think we could do this. Note, you can't require arguments in all places where there is a nested array selection, because there are no arguments on `__value` for a procedure result. So we have to pass a boolean along to track whether we expect that.

Also, looks like the spec pages are still missing. Can you please take a stab at writing something under `fields-selection.md`?

ndc-test/src/test_cases/query/common.rs
```diff
@@ -35,6 +46,7 @@ pub fn select_columns(
                 models::Field::Column {
                     column: f.0.clone(),
                     fields: None,
+                    arguments: None,
```
paf31, line 49: Now if the fields are nullable, we still need to provide them. It's probably easier to just exclude even fields with nullable arguments for now.

sordina: Aren't we doing this by having the arguments map optional? Otherwise I could just make the arguments map required regardless of what the arguments are.

paf31: Until we make (all) nullable inputs optional, for now, the individual arguments need to be provided as null, so this technically is an invalid request. Later we can just omit anything which is nullable.

sordina: Ok I've set arguments to null if they aren't required.

rfcs/0009-field-arguments.md
```diff
@@ -0,0 +1,57 @@
...
+}
+```
+
+and queries are extended with:
+
+```rust
+pub enum Field {
+  Column {
+    ...
+    arguments: Option<BTreeMap<String, Option<Argument>>>,
```
BenoitRanque, line 35: Could we implement this like so instead:
```rust
pub enum Field {
  Column {
    ...
    #[serde(skip_serializing_if = "BTreeMap::is_empty", default)]
    arguments: BTreeMap<String, Argument>,
  }
}
```
This maintains the same semantics when serialized while being a nicer API when de-serialized.

Also removed the inner option per the conversation elsewhere about arguments needing to specify null-ability rather than being optional.

ndc-models/src/lib.rs
```diff
@@ -205,6 +205,8 @@ pub struct ObjectField {
     /// The type of this field
     #[serde(rename = "type")]
     pub r#type: Type,
+    /// The arguments available to the field - Matches implementation from CollectionInfo
+    pub arguments: BTreeMap<String, ArgumentInfo>,
```
BenoitRanque, line 209: Per my other comment, consider:
```rust
#[serde(skip_serializing_if = "BTreeMap::is_empty", default)]
pub arguments: BTreeMap<String, ArgumentInfo>,
```

ndc-models/src/lib.rs
```diff
@@ -464,6 +466,7 @@ pub enum Field {
         /// by specifying fields to fetch here.
         /// If omitted, the column data will be fetched in full.
         fields: Option<NestedField>,
+        arguments: Option<BTreeMap<String, Argument>>,
```
BenoitRanque, line 469: Per my other comment consider:
```rust
#[serde(skip_serializing_if = "BTreeMap::is_empty", default)]
arguments: BTreeMap<String, Argument>,
```

sordina:
@paf31 I don't follow this block of code:

```rust
            let limit = if apply_array_arguments {
                let limit_argument = arguments
                    .as_ref()
                    .ok_or((
                        StatusCode::BAD_REQUEST,
                        Json(models::ErrorResponse {
                            message: "Expected arguments".into(),
ect...
```

I feel like this requires that limit is always supplied?

paf31:
> I feel like this requires that limit is always supplied?

Yes, I thought we agreed we'd make all listed arguments required and revisit this everywhere in a follow up PR.

ndc-test/src/test_cases/query/common.rs
paf31: `arguments` is still set to `None`? We need to either set it to a map with one entry for each argument, or we need to change the `matches!` macro to exclude any field with arguments (until the follow up PR).

sordina:
@paf31 

> Yes, I thought we agreed we'd make all listed arguments required and revisit this everywhere in a follow up PR.

Yes I agree with this, but it feels like "limit" specifically is being required if any arguments are required.

In the current state of the test suite this is the same thing, but if any other arguments are added this no longer makes sense.

sordina:
For example, it seems that if I added `where` too then and issued a query with a where field argument but not a limit, then it would error out because limit was missing.

sordina:
@codedmart The issue is that we're not checking if all arguments are supplied, we're just checking if "limit" is supplied.

ndc-models/src/lib.rs
sordina: 41518d9..7036c49

ndc-models/src/lib.rs
```diff
@@ -464,6 +466,7 @@ pub enum Field {
         /// by specifying fields to fetch here.
         /// If omitted, the column data will be fetched in full.
         fields: Option<NestedField>,
+        arguments: Option<BTreeMap<String, Option<Argument>>>,
```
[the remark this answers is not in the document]
sordina: 7036c49..ec562f5

ndc-test/src/test_cases/query/common.rs
sordina: 7036c49..ec562f5

rfcs/0009-field-arguments.md
sordina: 7036c49..ec562f5

ndc-models/src/lib.rs
sordina: 7036c49..ec562f5

ndc-test/src/test_cases/query/common.rs
```diff
@@ -1,19 +1,32 @@
...
+            if f.1.arguments.iter().all( |(_, v)|
+                matches!(v.argument_type, Type::Nullable { underlying_type: _ })
+            ) {
+               Some((
+                    f.0.clone(),
+                    models::Field::Column {
+                        column: f.0.clone(),
+                        fields: None,
+                        arguments: Some(f.1.arguments.keys().map(|k| {
+                            (k.to_owned(), None)
```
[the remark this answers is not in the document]
sordina: 7036c49..ec562f5

rfcs/0009-field-arguments.md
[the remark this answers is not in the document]
sordina: ec562f5..5779f24

specification/src/specification/queries/field-selection.md
```diff
@@ -20,6 +20,11 @@ If `fields` is provided, its value should be compatible with the type of the col
 - For an object-typed column (whether nullable or not), the `fields` property should contain a `NestedField` with type `object`. The `fields` property of the `NestedField` specifies a [`Field`](../../reference/types.md#field) structure for each requested nested field from the objects.
 - For an array-typed column  (whether nullable or not), the `fields` property should contain a `NestedField` with type `array`. The `fields` property of the `NestedField` should contain _another_ `NestedField` structure, compatible with the type of the elements of the array. The selection function denoted by this nested `NestedField` structure should be applied to each element of each array.
 
+Arguments can be supplied to `NestedField` objects via the `arguments` key. These match the format of [collection arguments](../arguments.md).
```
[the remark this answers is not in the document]
sordina: 032d538..4c65db8

ndc-models/tests/json_schema/mutation_request.jsonschema
```diff
@@ -141,19 +141,9 @@
               ]
             },
             "arguments": {
-              "type": [
-                "object",
-                "null"
-              ],
+              "type": "object",
```
paf31, line 144: This might be an issue - looks like `skip_serializing_if` doesn't work with `jsonschema`.

specification/src/specification/queries/field-selection.md
```diff
@@ -7,6 +7,28 @@ A [`Query`](../../reference/types.md#query) can specify which fields to fetch. T
...
 The requested fields are specified as a collection of [`Field`](../../reference/types.md#field) structures in the `field` property on the [`Query`](../../reference/types.md#query).
 
+## Field Arguments
+
+Arguments can be supplied to fields via the `arguments` key. These match the format described in [the arguments documentation](../arguments.md).
+
+The [schema response](../schema/object-types.md) will specify which fields take arguments via its respective `arguments` key.
+
+- If a field has any arguments specified, then all arguments must be provided when that field is queried
+- A literal argument `null` value may be used for optional arguments
```
paf31, line 17: Let's not use the word optional because it'll be confusing (required/optional vs nullable/non-nullable). We could say all arguments are required, including nullable arguments.

ndc-models/tests/json_schema/mutation_request.jsonschema
sordina: Oh. I'll see if @BenoitRanque knows a work around.

specification/src/specification/queries/field-selection.md
sordina:
```
- If a field has any arguments specified, then all arguments must be provided when that field is queried
- Literal argument `null` values may be used for nullable arguments
```

ndc-models/tests/json_schema/mutation_request.jsonschema
BenoitRanque: This is working as intended: `arguments` is [not a required field](https://github.com/hasura/ndc-spec/pull/139/files/7036c49c5941c88f2ab7ddefa9061455e6b4e667..ec562f57fb4bd073d480a3a7c26ff66f79487887#diff-9af8bfb06b0046ccdbcbe174bd4ee4d93320b43dc9ee9f8ce589d633858a0092L119-L120).

Do we additionally want to allow `arguments` to be present, but `null`?
That would leave 3 valid empty states: no `argument` field, `arguments` set to `null`, and `arguments` set to `{}`

sordina: @paf31 ?

specification/src/specification/queries/field-selection.md
sordina: Is this ok?

ndc-models/tests/json_schema/mutation_request.jsonschema
BenoitRanque: `schemars` picks up `#[serde(default)]` and makes the field optional. It's true it does not handle `skip_serializing_if`.

So, the schema validates either omitting the `arguments` key, or an object value.
Serde will additionally deserialize from a null value: both `null` and omitting the key are treated as the value being absent, and we fall back to the default value (in out case an empty map).

I do grant this is technically a bug in `schemars`. It looks like they [special cased](https://github.com/GREsau/schemars/issues/11) `Option` to be both nullable and optional.

I'd still argue to keep it this way, as the schema is still consistent with the serialized format.
We just happen to also be able to deserialize from misshapen payloads: null or empty arguments.

BenoitRanque: I must correct myself. `#[serde(default)]` seems to be special cased for `Option` and will happily serialize either null or missing fields to the default. But it will error for other types. So the schemars behavior is correct.

So with the annotations above, `"arguments": null` is not valid, and the schema reflects that.

Do we absolutely need to be able to serialize from `"arguments": null` ?

paf31: Looks good to me. The tests confirm it's backwards compatible, which is the main thing.

ndc-reference/bin/reference/main.rs
```diff
@@ -1878,10 +1908,47 @@ fn eval_nested_field(
                     }),
                 )
             })?;
-            let result_array = array
-                .into_iter()
+
+            let limit = if apply_array_arguments {
```
paf31, line 1912: There is a bug here: if `fields` is not provided, we never reach this case, so we don't apply the limit. We need to lift this logic out into `eval_field`.

paf31:
I spotted a couple of bugs and I think I've fixed them, but I need to review. The new code also needs some tidying:

- `ndc-test` didn't exclude non-nullable columns in `select_columns`. That code now needs deduplicating.
- `ndc-reference` didn't parse arguments in the right place, leading us to not apply `limit` unless nested fields were requested.

Also, I noticed (and documented) that we can't support field arguments in column mappings for relationships, or to define relationship arguments, or in nested field path selectors. I think that'll be fine for now but I want to make sure that's ok @codedmart.

sordina:
I think this incorrectly always requires a limit arg:

```rust
// ANCHOR_END: eval_comparison_target
// ANCHOR: eval_column
fn eval_column(
    variables: &BTreeMap<String, serde_json::Value>,
    row: &Row,
    column_name: &str,
    arguments: &BTreeMap<String, models::Argument>,
) -> Result<serde_json::Value> {
    let column = row.get(column_name).cloned().ok_or((
        StatusCode::BAD_REQUEST,
        Json(models::ErrorResponse {
            message: "invalid column name".into(),
            details: serde_json::Value::Null,
        }),
    ))?;

    if let Some(array) = column.as_array() {
        let limit_argument = arguments.get("limit").ok_or((
            StatusCode::BAD_REQUEST,
            Json(models::ErrorResponse {
                message: "Expected argument 'limit'".into(),
                details: serde_json::Value::Null,
            }),
        ))?;
```

```
failures:

---- tests::test_query stdout ----
thread 'tests::test_query' panicked at ndc-reference/bin/reference/main.rs:2416:22:
called `Result::unwrap()` on an `Err` value: (400, Json(ErrorResponse { message: "Expected argument 'limit'", details: Null }))
```

sordina:
which makes having backwards compatibility impossible?

paf31 (approved):

PR #140, opened 2024-05-07
Upgrade Rust to v1.78.0, and the Rust dependencies to their latest versions.

SamirTalwar:
Rust v1.77.2 contains a security upgrade, so we should use it or a later version (I have chosen the latest, v1.78.0) when building the ndc-test program for release.

I also upgraded all dependencies in _Cargo.lock_ so that if there are any security upgrades there, we include them, but just used the lowest semver-compatible version for the _Cargo.toml_ files to ensure that it's easy to pull in e.g. ndc-models without having to bump everything.

`rustc` v1.78.0 detected some dead code so I also removed it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,47 +43,48 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.13"
+version = "0.6.14"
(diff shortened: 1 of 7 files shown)


paf31 (changes requested):
Thanks!

paf31 (approved):

PR #141, opened 2024-05-13
ndc-test: validate response types

paf31:
- [x] Support this in `replay` mode
- [x] Put this behind a flag


ndc-test/src/error.rs
```diff
@@ -65,6 +65,10 @@ pub enum Error {
     BenchmarkExceededTolerance(f64),
     #[error("response from connector does not satisfy requirement: {0}")]
     ResponseDoesNotSatisfy(String),
+    #[error("invalid response at path {}: expected {1}", .0.join("."))]
```
codedmart, line 68: `.0.join(".")` this is interesting. I am assuming `0` is a string and then you join with a `.`, but why the extra `.` at the start?

codedmart (approved):

ndc-test/src/error.rs
paf31: `.0` is just `#[error]` syntax for "the first argument", so this is "take the first argument (an array) and join it with `.`".

codedmart (approved):

PR #143, opened 2024-05-29
Add impl Default for NestedFieldCapabilities

paf31:
This is to support JSON backwards-compatibility.

### DIFF
--- a/ndc-models/src/lib.rs
+++ b/ndc-models/src/lib.rs
@@ -61,6 +61,7 @@ pub struct QueryCapabilities {
     /// Does the connector support explaining queries
     pub explain: Option<LeafCapability>,
     /// Does the connector support nested fields
+    #[serde(default)]
     pub nested_fields: NestedFieldCapabilities,
 }
 // ANCHOR_END: QueryCapabilities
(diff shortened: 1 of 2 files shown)


ndc-models/src/lib.rs
```diff
@@ -61,6 +61,7 @@ pub struct QueryCapabilities {
     /// Does the connector support explaining queries
     pub explain: Option<LeafCapability>,
     /// Does the connector support nested fields
+    #[serde(default)]
     pub nested_fields: NestedFieldCapabilities,
```
abhinav-hasura, line 65: Should this be `Option<NestedFieldCapabilities>`, where `None` would mean, no nested field support is available?

paf31: No, because "no nested field support" can be indicated by not returning nested types in the schema response. Currently, if you declare nested fields, we assume you can at least select them, but we could add a sub-capability for that too.

abhinav-hasura (dismissed):

SamirTalwar (dismissed):

SamirTalwar:
Can we make a new release (0.13.1?) so we can unblock releasing connectors?

SamirTalwar:
I meant to say 0.1.4, of course. #145 will need to be merged and then we can release.

PR #145, opened 2024-06-04
Bump the version to 0.1.4.

SamirTalwar:


soupi (approved):

ndc-reference/bin/reference/main.rs
```diff
@@ -2353,7 +2353,7 @@ mod tests {
 
             let mut expected = mint.new_goldenfile(expected_path).unwrap();
 
-            let response_json = serde_json::to_string_pretty(&response.0).unwrap();
+            let response_json = serde_json::to_string_pretty(&response.0).unwrap() + "\n";
```
paf31, line 2356: Is this deliberate?

SamirTalwar:
I'm putting this into draft until we decide it's a good idea to release a new version.

ndc-reference/bin/reference/main.rs
paf31: We shouldn't be editing these by hand really, though. You can run `UPDATE_GOLDENFILES=1 cargo test` to update golden files when there is a breaking change.

SamirTalwar: I have had trouble getting that command to work. (I highly recommend we switch to [Insta](https://insta.rs/).)

PR #146, opened 2024-06-05
0.2.0 release branch - ComparisonTarget changes, add named scopes, group by and filter by aggregates, nested collections

paf31:
- [x] RFC
  - [x] [Remove `ComparisonTarget["path"]` RFC](https://github.com/hasura/ndc-spec/blob/phil/remove-comparisontarget-path/rfcs/0011-no-paths-in-comparison-target.md)
  - [x] [Named scopes RFC](https://github.com/hasura/ndc-spec/blob/phil/remove-comparisontarget-path/rfcs/0015-named-scopes.md)
  - [x] [Group by RFC](https://github.com/hasura/ndc-spec/blob/phil/remove-comparisontarget-path/rfcs/0014-group-by.md)
  - [x] [Filter by aggregates RFC](https://github.com/hasura/ndc-spec/blob/phil/remove-comparisontarget-path/rfcs/0013-filter-by-aggregates.md)
  - [x] [Nested collections RFC](https://github.com/hasura/ndc-spec/blob/phil/remove-comparisontarget-path/rfcs/0019-nested-collections.md)
- [x] Specification updates
  - [x] Changelog
  - [x] Specification pages
  - [x] Tutorial pages
  - [x] `reference/types.md`


daniel-chambers:
Looking at usages of `ComparisonTarget`'s `path` property in v3-engine right now, it's never used (ie it is set to `[]`). So it would be safe to delete it and we wouldn't lose any features.

soupi:
This will be a breaking change because the `path` field is not optional. Meaning that existing connectors will not work with v3-engine versions that include this change.

BenoitRanque:
Sorry accidental enter closed the issue ☹️

paf31:
> I'm not against the proposal in this RFC, as GDC never supported what NDC allows with column path, however, without the scopes stuff specified in "future work", we can't support the same feature in Hasura v3.

@daniel-chambers We'll still have `ComparisonTarget::RootCollectionColumn` for now, even before the scopes capability comes in (at which point it will be redundant). We could consider removing that now too, but I figured keep it for parity with v2, even if we don't use it right away. I'm not sure what I prefer though, honestly. If we're not using it in engine yet, might be better to rip it out and deal with scopes all at once.

BenoitRanque:
Has anyone actually implemented root collection column comparisons? I know I haven't.
But if some connectors do have code to handle that use case, maybe removing it would mean they'd have to remove the code, and eventually add it back in when scopes land. Which would be more annoying than migrating to scopes from root collection column.

daniel-chambers:
> This will be a breaking change because the path field is not optional. Meaning that existing connectors will not work with v3-engine versions that include this change.

Ugh, this is a problem. Terrible idea: could we hack the Rust types to continue serializing a property with a default `[]` and remove the actual property from the types?

> We'll still have ComparisonTarget::RootCollectionColumn for now, even before the scopes capability comes in (at which point it will be redundant) [...] If we're not using it in engine yet, might be better to rip it out and deal with scopes all at once.

@paf31 As far as I can tell, we're not using it in the engine yet (a search across the codebase only finds a usage in the custom connector used for testing, not in the engine itself). I'd be for ripping it out now before we end up with two ways to achieve the same thing in the spec and stuck with them.

soupi:
@BenoitRanque this has been implemented in [ndc-postgres](https://github.com/hasura/ndc-postgres/blob/main/crates/query-engine/translation/src/translation/query/filtering.rs#L310), yes. And it is not a trivial bit of code. If we don't plan to use it ever, then ripping it out will make things simpler, but I wouldn't want to rip it out only to re-implement it in the future.

danieljharvey:
> And it is not a trivial bit of code. If we don't plan to use it ever, then ripping it out will make things simpler, but I wouldn't want to rip it out only to re-implement it in the future.

Although this is frustrating, there is also the ongoing cost of keeping this code maintained and tested to consider, a job which is trickier when the code paths aren't ever really exercised for real.

paf31:
I know there's a cost to reshaping code that already exists, but it's a one-time cost, and I think the cost might be greater if we wait, since we plan to add more features which might have to work together with whatever we choose here.

codedmart:
@hallettj Can you weigh in here please. Have you implemented this in Mongo?

hallettj:
I was just working on making binary comparisons work where both sides of the comparison are targets with paths. It turns out to be obnoxiously complicated. Tbh I would be relieved to be able to simplify that to only having a path on the RHS.

I'm also still working on root collection column references. I think I have to do something like the named scopes idea internally in the mongodb connector anyway. I'd be fine with replacing root column references with scopes.

According to the current [docs](https://hasura.github.io/ndc-spec/specification/queries/filtering.html?highlight=Exists#referencing-a-column-from-the-root-collection),

> The root collection is defined as the collection in scope at the nearest enclosing [Query](https://hasura.github.io/ndc-spec/reference/types.html#query)

That means that each related field introduces a new "root", right? I think scopes could make that more clear. I've had some conversations already along the lines of, "Wait, the root collection doesn't mean the collection entrypoint to the entire query request?"

paf31:
> That means that each related field introduces a new "root", right?

Yes, any time you start a new `Query` (basically just `Field::Relationship` right now), the root gets reset to the current table. Within that `Query`, the scope can change whenever you use `Exists`, and the named scope stack would track all scopes between the current one and the current root.

paf31:
I would like to consider bundling this with the following breaking changes:

- Remove `RootCollectionColumn`. It really belongs in `ComparisonValue` anyway, but it would be unnecessary after the named scopes feature anyway.
- Change `ComparisonValue` to refer to columns directly, instead of via `ComparisonTarget`. This way we can extend `ComparisonTarget` with more things which should only appear on the left hand side of a comparison (like aggregates).
- (Optional) removing one or the other of `ExistsInCollection::Related` or `ExistsInCollection::Unrelated`. With named scopes, each can be expressed in terms of the other. `ExistsInCollection::Related` gives a way to compress the request by referring to the same relationship only by name, but `ExistsInCollection::Unrelated` is arguably the more primitive of the two. However, using `ExistsInCollection::Unrelated` everywhere forces the caller to re-specify relationships using named scopes, which would be tedious.

paf31:
@daniel-chambers About `RootCollectionColumn`:

> I'd be for ripping it out now before we end up with two ways to achieve the same thing in the spec and stuck with them.

The tricky thing is that this makes exists-unrelated basically useless (for now) so I think we either need to consolidate both types of exists, or we need to implement named scopes now. Or we can remove exists-unrelated for now and add it back later if/when we need it.

daniel-chambers:
> The tricky thing is that this makes exists-unrelated basically useless (for now)

@paf31 As far as I can tell, exists unrelated is unused in v3-engine at the moment, so if we know it's going to get used by named scoped in the future, we could safely leave it in even if it is unused, so as to not incur a breaking change on connectors only to reintroduce the feature in the future.

> (Optional) removing one or the other of ExistsInCollection::Related or ExistsInCollection::Unrelated.

Yes, you can express `ExistsInCollection::Related` in terms of `ExistsInCollection::Unrelated` + named scopes, but I don't see that as a compelling reason to throw out `ExistsInCollection::Related`. Exists involving relationships will get expressed as `ExistsInCollection::Related` because its convenient and fits the existing model, and `ExistsInCollection::Unrelated` will be used for other sorts of queries (with or without using named scopes). I think I prefer the retaining the consistency of how relationships are handled in NDC, at the small cost of a spot where you could express the same semantics in two distinct ways.

> With named scopes, each can be expressed in terms of the other.

I'm not sure how `ExistsInCollection::Unrelated` could be expressed using `ExistsInCollection::Related` + named scopes, unless you're talking about defining a relationship with zero `column_mapping`s, which would be super weird. That almost feels like an abuse of the relationships system. I'm not sure that unintuitive usage is worth it to remove the overlap in functionality.

ndc-reference/tests/query/predicate_with_unrelated_exists_and_relationship/request.json
```diff
@@ -38,43 +38,22 @@
                     "predicate": {
                         "type": "exists",
                         "in_collection": {
-                            "type": "unrelated",
+                            "type": "related",
```
daniel-chambers, line 41: This test might need a renaming given that it doesn't actually do any unrelated exists stuff anymore.

daniel-chambers (approved):
LGTM, with one suggestion.

ndc-reference/tests/query/predicate_with_unrelated_exists_and_relationship/request.json
paf31: Fixed, thanks!

daniel-chambers (approved):

PR #147, opened 2024-06-06
Filter by aggregates

paf31:
[Rendered RFC](https://github.com/hasura/ndc-spec/blob/filter-by-aggregate/rfcs/0013-filter-by-aggregates.md)

---

- [x] RFC
- [x] Specification updates
  - [ ] Changelog
  - [x] Specification pages
  - [x] Tutorial pages
  - [x] `reference/types.md`


paf31:
One concern here is that aggregates can now also appear on the right-hand side of a comparison operator too. I think we generally need to rethink what can go in `ComparisonTarget` vs `ComparisonValue`, and whether the two need to just be completely separated.

daniel-chambers:
> One concern here is that aggregates can now also appear on the right-hand side of a comparison operator too. 

Hrm. It _would_ be preferable to prevent this until this becomes a feature that we actually want. Right now, as far as I'm concerned, it's very outside of scope. I can see that theoretically someone might want it, but I'd rather wait for a real customer use case.

paf31:
@daniel-chambers This is now rebased onto the breaking changes branch. It includes column counts (via a breaking change to use `Aggregate` in both filter and order-by), and avoids aggregates in `ComparisonValue`s.

paf31:
@daniel-chambers I realized there is an issue with this feature. Single column aggregates have a declared scalar return type, which means they have specific comparison operators which ought to work correctly with aggregates of that type. However, `column_count` and `star_count` aggregates do not. They return integers (as JSON), but not any particular scalar type, so there is no way currently to say what operators ought to exist.

A few options:

1. Have the caller declare the intended return type in the IR (gross)
2. Break the schema response to declare the types there (probably my preferred option)
3. Break the query request IR now, remove `column_count` and have it be just another aggregate operator with a `distinct` argument (doesn't solve for `star_count` though)

daniel-chambers:
@paf31 Good point. My preference is also number 2.

paf31:
@daniel-chambers I've added a data structure to the schema response which mirrors the capabilities response, so that we can add schema-level data which is required by certain capabilities. I think this is ready for a review if you don't mind.

daniel-chambers (approved):

PR #148, opened 2024-06-06
Feature: Group by

paf31:
[Rendered RFC](https://github.com/hasura/ndc-spec/blob/feature/group-by/rfcs/0014-group-by.md)

---

- [x] RFC
- [x] Specification updates
  - [ ] Changelog
  - [x] Specification pages
  - [x] Tutorial pages
  - [x] `reference/types.md`
  - [x] Check the capability in `ndc-test`


ndc-models/src/lib.rs
```diff
@@ -689,9 +717,25 @@ pub struct RowSet {
...
+#[skip_serializing_none]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Group")]
+pub struct Group {
+    /// 
+    pub dimensions: IndexMap<String, serde_json::Value>,
+    ///
+    pub aggregates: IndexMap<String, serde_json::Value>,
+    ///
+    pub rows: Vec<IndexMap<String, RowFieldValue>>,
```
daniel-chambers, line 735: We removed this from scope. You can't return rows of groups in the current Aggregates/Grouping RFC.

ndc-models/src/lib.rs
```diff
@@ -420,9 +420,37 @@ pub struct Query {
...
+// ANCHOR_END: Grouping
+
+// ANCHOR: Dimension
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[skip_serializing_none]
+#[serde(tag = "type", rename_all = "snake_case")]
+#[schemars(title = "Dimension")]
+pub enum Dimension {
+    Column {
+        column_name: String,
```
daniel-chambers, line 449: We'd need a column path thing here to address grouping over nested object fields.

ndc-models/src/lib.rs
```diff
@@ -420,9 +420,37 @@ pub struct Query {
...
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Grouping")]
+pub struct Grouping {
+    /// Dimensions along which to partition the data
+    pub dimensions: IndexMap<String, Dimension>,
+    /// Fields to fetch in each group
+    pub fields: IndexMap<String, Field>,
+    /// Aggregates to compute in each group
+    pub aggregates: IndexMap<String, Aggregate>,
+}
```
daniel-chambers, line 439: I think we also need to add `having`, `limit`, `offset` and group `order_by` concepts here too.

ndc-models/src/lib.rs
```diff
@@ -420,9 +420,37 @@ pub struct Query {
...
 }
 // ANCHOR_END: Query
 
+// ANCHOR: Grouping
+#[skip_serializing_none]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Grouping")]
+pub struct Grouping {
+    /// Dimensions along which to partition the data
+    pub dimensions: IndexMap<String, Dimension>,
```
daniel-chambers, line 434: I'm concerned about using a JSON object to represent this structure. The ordering of dimensions is potentially important (see Rollup grouping type in the RFC; yes this is out of scope for now, but let's not ignore the potential) and JSON objects don't care about property ordering.

ndc-models/src/lib.rs
paf31: `having` is the only one that's new, right? I have a separate JIRA issue for that.

The others are specified already on the `Query` level, unless you mean sorting and paginating the _groups_ themselves, but I don't think that's a common pattern?

ndc-models/src/lib.rs
paf31: `IndexMap` does preserve the order of keys, so you can consider it as an ordered map structure. But maybe it's better to change this to a vector. I would then also remove the names from the query request.

paf31: Note that the order of fields in the query response potentially is also important, which is why we use `IndexMap` (for GraphQL at least, it should match the requested order)

ndc-models/src/lib.rs
```diff
@@ -689,9 +717,25 @@ pub struct RowSet {
...
 }
 // ANCHOR_END: RowSet
 
+// ANCHOR: Group
+#[skip_serializing_none]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Group")]
+pub struct Group {
+    /// 
+    pub dimensions: IndexMap<String, serde_json::Value>,
```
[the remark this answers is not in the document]
paf31: Yes, it just returns all selected dimensions.

ndc-models/src/lib.rs
paf31: Are we taking about ordering and paginating the _groups_ themselves, e.g. order by count-of-group? So then we'd have two order by's, two limits, and two offsets?

daniel-chambers: > Are we taking about ordering and paginating the groups themselves, e.g. order by count-of-group? So then we'd have two order by's, two limits, and two offsets?

Yes.

You can filter, order and paginate the input to your grouping ([see here](https://github.com/hasura/v3-engine/blob/main/rfcs/aggregations.md#filtering-ordering-and-paginating-the-objects-grouped-over)), and then after the grouping and group aggregates are performed over those input rows, you can filter the groups (having), order them, and paginate them.

Here's a fairly nonsensical example, but it uses all the features:

```graphql
  query MyQuery {
    Invoice_groups(
      filter_input: {
        where: {BillingAddress: {Country: {_eq: "Australia"}}}, 
        order_by: {BillingAddress: {State: Asc}}, 
        offset: 1, 
        limit: 10
      }
      grouping_keys: {Customer: {_scalar_field: LastName}}
      having: {_count: {_gt: 2}}
      order_by: {group_key: {Customer: {FirstName: Asc}}}
      limit: 5
      offset: 1
    ) {
      group_key {
        Customer {
          LastName
        }
      }
      group_aggregate {
        _count
      }
    }
  }

```
(play with it [here](https://cloud.hasura.io/public/graphiql?endpoint=https%3A%2F%2Fhasura-v3-aggregates-mock.deno.dev&query=query+MyQuery+%7B%0A++Invoice_groups%28%0A++++filter_input%3A+%7B%0A++++++where%3A+%7BBillingAddress%3A+%7BCountry%3A+%7B_eq%3A+%22Australia%22%7D%7D%7D%2C+%0A++++++order_by%3A+%7BBillingAddress%3A+%7BState%3A+Asc%7D%7D%2C+%0A++++++offset%3A+1%2C+%0A++++++limit%3A+10%0A++++%7D%0A++++grouping_keys%3A+%7BCustomer%3A+%7B_scalar_field%3A+LastName%7D%7D%0A++++having%3A+%7B_count%3A+%7B_gt%3A+2%7D%7D%0A++++order_by%3A+%7Bgroup_key%3A+%7BCustomer%3A+%7BFirstName%3A+Asc%7D%7D%7D%0A++++limit%3A+5%0A++++offset%3A+1%0A++%29+%7B%0A++++group_key+%7B%0A++++++Customer+%7B%0A++++++++LastName%0A++++++%7D%0A++++%7D%0A++++group_aggregate+%7B%0A++++++_count%0A++++%7D%0A++%7D%0A%7D%0A))

ndc-models/src/lib.rs
daniel-chambers: My issue is less about the data structure chosen in Rust, and more that it translates to a JSON object in the JSON Schema. JSON Schema doesn't say anything about object property ordering being retained afaik. And if other programming languages (cough, Haskell with `aeson`) consume JSON Objects they often shove them in hashmaps and the ordering is lost.

If the ordering is important, we should serialize to an array in JSON IMHO, so that's explicit.

ndc-models/src/lib.rs
paf31: I've added `HAVING` to this PR, but I haven't finished order by yet.

rfcs/0014-group-by.md
```diff
@@ -0,0 +1,97 @@
...
+The exact semantics of this partitioning operation are unspecified. For example, one connector might implement sort-then-group, and another might choose not to sort and return multiple groups for the same dimensions instead.
+
+## Future Work
+
+- Support ordering and filtering after the grouping operation
+  - Post-group filtering is called `HAVING` in SQL.
+
+## Open Questions
+
+- A `Grouping` with no dimensions is just a regular aggregation, do we want to unify the two?
```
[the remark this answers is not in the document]
paf31: I ended up nesting the capability under `aggregations`, given that one implies the other, but having thought a bit more, I do like having them be separable in the IR if a connector supports one but not the other.

rfcs/0014-group-by.md
```diff
@@ -0,0 +1,97 @@
...
+- Do we want to support multiple grouping operations in the same query?
```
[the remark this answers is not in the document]
paf31: Can't the user just request groups twice, underneath two separate aliased fields? Presumably we'd want that to be run as a single query if we can.

ndc-models/src/lib.rs
paf31: Agreed, I've changed the dimensions list to be a vector.

ndc-models/src/lib.rs
```diff
@@ -420,9 +420,37 @@ pub struct Query {
...
 
+// ANCHOR: Grouping
+#[skip_serializing_none]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Grouping")]
+pub struct Grouping {
+    /// Dimensions along which to partition the data
+    pub dimensions: IndexMap<String, Dimension>,
+    /// Fields to fetch in each group
+    pub fields: IndexMap<String, Field>,
```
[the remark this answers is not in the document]
paf31: I've dropped this, but remind me why it was removed? I think there's a good argument for a drill-down operation to be supported easily by adding a field to your GraphQL query. 

And it would be more reliable to fetch rows per-group rather than to expect the engine to get the grouping right.

A related possible feature is to express relationships from dimension columns. For example, if I group articles by `author_id`, it's sensible to want to be able to fetch the related author, once-per-group and not once-per-row.

ndc-models/src/lib.rs
paf31: Order-by is here now as well.

ndc-models/src/lib.rs
```diff
@@ -420,9 +440,72 @@ pub struct Query {
...
+        expressions: Vec<GroupExpression>,
+    },
+    Not {
+        expression: Box<GroupExpression>,
+    },
+    AggregateComparison {
+        aggregate: Aggregate,
+        operator: String,
+        value: AggregateComparisonValue,
+    },
```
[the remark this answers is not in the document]
paf31: I think this is now much more similar to the other branch, although I haven't rebased this branch yet, so it's hard to see.

ndc-models/src/lib.rs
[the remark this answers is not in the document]
daniel-chambers: The argument was that if you cared about row data, you could just pull back the rows and group yourself client side. There's very little advantage having the server do the grouping for you, since you'd be pulling all row data anyway.

Aggregates over groups are different as they don't require pulling row data to the client, and can be optimised by the server.

rfcs/0014-group-by.md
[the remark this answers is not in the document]
daniel-chambers: Two separate aliased root fields are still two separate queries to NDC right now.

Am I missing the point you're trying to make? Perhaps an example in GraphQL would help.

ndc-models/src/lib.rs
```diff
@@ -420,11 +440,111 @@ pub struct Query {
...
+#[schemars(title = "Grouping")]
+pub struct Grouping {
+    /// Dimensions along which to partition the data
+    pub dimensions: Vec<Dimension>,
+    /// Aggregates to compute in each group
+    pub aggregates: IndexMap<String, Aggregate>,
+    /// Optionally specify a predicate to apply after grouping rows
+    pub predicate: Option<GroupExpression>,
+    /// Optionally specify how rows should be ordered
+    pub order_by: Option<GroupOrderBy>,
```
daniel-chambers, line 464: Still need a limit and offset in here too.

ndc-models/src/lib.rs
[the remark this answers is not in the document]
daniel-chambers: It's more similar now, except the difference I raised above about binary and unary operators. What are you thoughts about that?

ndc-models/src/lib.rs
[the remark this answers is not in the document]
paf31: I don't think it makes sense to have the engine try to replicate the behavior of the database. If we group by a `citext` column for example, is the engine going to do the right thing? Also, that implies that the semantics of grouping are fixed, but maybe they shouldn't be. Sorted grouping and unsorted grouping are both valid IMO, for example.

Finally, there maybe be a future concrete benefit to doing this server side: as I say, you can fetch less data if you can express relationships from dimension columns.

rfcs/0014-group-by.md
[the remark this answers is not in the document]
paf31: Let's say I want to group the same rows two different ways. Then I need to make two queries, but a mutation might run in between them, which would make the combined result invalid.

On the other hand, I can fetch the same field twice with different arguments, say, in the same query.

ndc-models/src/lib.rs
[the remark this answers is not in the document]
paf31: I agree about unary operators (although I think testing for `SUM(..) IS NULL` is a bit weird to be honest, and I can't think of a good use case), but I didn't see anything about binary operators. Did you mean that I should separate out an `AggregateComparisonTarget` with only one variant?

ndc-models/src/lib.rs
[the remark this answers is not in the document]
daniel-chambers: > I don't think it makes sense to have the engine try to replicate the behavior of the database

Nobody has suggested the v3-engine should be doing this. The GraphQL API user would do it.

> you can fetch less data if you can express relationships from dimension columns.

Yeah, that's a valid point; or at least you can not do two queries on the GraphQL API client end. One where you query the dimensions and one where you perform another query to load the related entities. To clarify, are you saying then that fields are only useful for dimension columns, or do you still think loading group row data is necessary. They seem separate to me.

Is there any reason we couldn't add this functionality incrementally in the future if customers asked for it, rather than add it now?

rfcs/0014-group-by.md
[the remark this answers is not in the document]
daniel-chambers: I guess my point is that this is fundamentally a problem with NDC as a whole right now, not just grouping. If I want to query the same table twice with different filters inside the same transaction, I can't. If I want to perform aggregations against the same table, but with a different input filter, inside the same transaction, I can't.

It seems to me that this could be solved more generally by allowing multiple query requests to be sent and executed inside the same transaction, rather than creating a special case for grouping, when from the graphql perspective all of these different query types that I may want to group together look the same (ie. a single graphql query with multiple root fields, aliased or otherwise).

ndc-models/src/lib.rs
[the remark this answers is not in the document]
daniel-chambers: I'm not quite sure what you mean, but the changes you've made solve the problem I raised. 🙂

I agree with your assessment of the usefulness of `SUM(..) IS NULL`. I guess I was kinda futureproofing in case there were unary operators that I didn't know about from some weird database that did actually make sense. Personally I don't think it hurts to have the support there, especially since we support them in regular predicates, but I wouldn't object if you wanted to remove unary operator support here, but leave the current structure in place so that it would be easy to incrementally add it back again if we found a real use case for it.

ndc-models/src/lib.rs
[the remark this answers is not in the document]
daniel-chambers: To clarify, I see this as useful:

```graphql
query MyQuery {
  Invoice_groups(grouping_keys: {Customer: {_scalar_field: CustomerId}}) {
    group_key {
      Customer {
        CustomerId # This is the actual group key
        Address { # Load the related address!
          City
          Country
          PostalCode
          State
          StreetAddress
        }
      }
    }
  }
}
```

but it doesn't translate to SQL very cleanly, because technically all the related Address fields need to be added to the grouping key for this to work (and it would work right now, if you did add them, no?).

This is not so useful, because you end up just getting the Customer Address just repeated over and over for each group row, rather than once per customer group:

```graphql
query MyQuery {
  Invoice_groups(grouping_keys: {Customer: {_scalar_field: CustomerId}}) {
    group_key {
      Customer {
        CustomerId # This is the actual group key
      }
    }
    group_rows { # This is not in the RFC
      Customer {
        Address { # Load the related address!
          City
          Country
          PostalCode
          State
          StreetAddress
        }
      }
   }
  }
}
```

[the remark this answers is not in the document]
daniel-chambers: Actually, looking at the type of `Dimension` again, we actually can't express the first graphql query at all, since we can't address object-related columns. Perhaps this needs to change and then we could achieve what you're talking about like this:

```
query MyQuery {
  Invoice_groups(
    grouping_keys: [ 
      {Customer: {_scalar_field: CustomerId}},
      {Customer: {Address: {_scalar_field: StreetAddress}}},
      {Customer: {Address: {_scalar_field: City}}},
      {Customer: {Address: {_scalar_field: Country}}},
      {Customer: {Address: {_scalar_field: PostalCode}}},
      {Customer: {Address: {_scalar_field: State}}}
    ]
  ) {
    group_key {
      Customer {
        CustomerId
        Address {
          City
          Country
          PostalCode
          State
          StreetAddress
        }
      }
    }
  }
}
```

ndc-test/src/test_cases/query/grouping/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+
+                connector.query(query_request.clone()).await?;
+            }
+        }
+
+        Ok(())
+    })
+}
+
+// pub async fn test_star_count_aggregate<C: Connector>(
```
daniel-chambers, line 72: How come all this is commented out?

ndc-models/src/lib.rs
```diff
@@ -691,9 +832,23 @@ pub struct RowSet {
...
 }
 // ANCHOR_END: RowSet
 
+// ANCHOR: Group
+#[skip_serializing_none]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Group")]
+pub struct Group {
+    /// Values of dimensions which identify this group
+    pub dimensions: Vec<serde_json::Value>,
```
soupi, line 846: Why is this not an `IndexMap<String, serde_json::Value>`?

ndc-reference/tests/query/group_by_with_star_count/expected.json
```diff
@@ -0,0 +1,22 @@
...
+  {
+    "groups": [
+      {
+        "dimensions": [
+          1
+        ],
+        "aggregates": {
+          "article_count": 1
+        }
+      },
```
soupi, line 11: I'm wondering why this isn't:

```json
      {
        "author_id": 1,
        "article_count": 1
      },
```

?

ndc-models/src/lib.rs
paf31: This is how it was before, but @daniel-chambers made the good point that the ordering of dimensions may be significant, and JSON at the level of values does not respect the ordering of fields.

ndc-reference/tests/query/group_by_with_star_count/expected.json
paf31: Because dimensions don't have names, and if they did, they might overlap.

rfcs/0014-group-by.md
```diff
@@ -0,0 +1,97 @@
...
+- The returned groups should correspond to a partition of the row set computed from the query:
+  - Within each partition, the chosen dimensions must be equal. These common values are hoisted out to the `dimensions` property of the `Group`.
+  - `aggregates` are computed over each partition in turn, in the same way as the aggregates API would compute them over the selected rows.
+
+The exact semantics of this partitioning operation are unspecified. For example, one connector might implement sort-then-group, and another might choose not to sort and return multiple groups for the same dimensions instead.
+
+## Future Work
+
+- Support ordering and filtering after the grouping operation
+  - Post-group filtering is called `HAVING` in SQL.
```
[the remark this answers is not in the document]
paf31: Thanks, I originally added this to the aggregates module and then split it out.

ndc-test/src/test_cases/query/grouping/mod.rs
paf31: Thanks, I originally added this to the aggregates module and then split it out.

ndc-models/src/lib.rs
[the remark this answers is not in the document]
paf31: > Is there any reason we couldn't add this functionality incrementally in the future if customers asked for it, rather than add it now?

Yes, we can add it later, I just wanted to make sure I understood the decision.

ndc-models/src/lib.rs
```diff
@@ -420,11 +444,128 @@ pub struct Query {
...
+#[skip_serializing_none]
+#[serde(tag = "type", rename_all = "snake_case")]
+#[schemars(title = "Dimension")]
+pub enum Dimension {
+    Column {
+        /// The name of the column
+        column_name: String,
+        /// Path to a nested field within an object column
+        field_path: Option<Vec<String>>,
+    },
```
daniel-chambers, line 532: Could we add a relationship path, so we could support grouping by object-related fields? (As discussed [here](https://github.com/hasura/ndc-spec/pull/148#discussion_r1646814189))

ndc-reference/tests/query/group_by_with_star_count/expected.json
soupi: Could you give an example please? My context for understanding this is sql, I understand that dimensions are effectively group by keys? And group by keys are named. Am I missing something?

soupi: Also, am I right in understanding that we have a separation between dimensions and aggregates in the result to mimic the request structure?

soupi (approved):

ndc-reference/tests/query/group_by_with_star_count/expected.json
paf31: The dimensions/aggregates split comes from multidimensional analysis. It's a bit awkward in SQL because everything is combined into one expression language, but then not all expressions are equal - you can't compute all expressions per-group, for example. In SQL, everything would be a named member of the target list, but in general, you just have an ordered list of dimensions to partition the data by. 

Yes, the response is the same as the request.

rfcs/0014-group-by.md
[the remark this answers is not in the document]
paf31: Yes, that's true. Not worth worrying about here, we can address this more generally later.

ndc-models/src/lib.rs
paf31: Done, thanks!

rfcs/0014-group-by.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+Requirements:
+
+- If `groups` is specified, then a `groups` property should also be provided in each row set returned.
+- The returned groups should correspond to a partition of the row set computed from the query:
+  - Within each partition, the chosen dimensions must be equal. These common values are hoisted out to the `dimensions` property of the `Group`.
+  - `aggregates` are computed over each partition in turn, in the same way as the aggregates API would compute them over the selected rows.
+
+The exact semantics of this partitioning operation are unspecified. For example, one connector might implement sort-then-group, and another might choose not to sort and return multiple groups for the same dimensions instead.
```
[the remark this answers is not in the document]
paf31: True, I can't think of a SQL database which groups like this, but I don't think there's any harm in keeping it slightly generalized for now. We can always specify it more tightly later if we decide we need this specific behavior.

But I can imagine some data stores grouping without sorting. For streaming data, for example, it's natural to group in batches, so I could imagine it coming up. It's also common for iterator libraries to provide both options.

paf31:
@soupi @daniel-chambers I think this is now ready for final review. I've rebased it onto #146, which has become the 0.2.0 release branch, essentially.

specification/src/specification/queries/grouping.md
```diff
@@ -0,0 +1,83 @@
...
+## Pagination
+
+Pagination can also be applied both before and after grouping:
+
+- The initial row set can be paginated _before the grouping operation_, using the `limit` and `offset` fields of the [`Query`](../../reference/types.md#query) object as usual, and
+- The _groups themselves_ can be paginated _after the grouping operation_, using the `limit` and `offset` fields of the [`Grouping`](../../reference/types.md#grouping) object. This is controlled by the `queries.aggregates.group_by.paginate` capability.
+
+## Examples
+
+This example paritions the `articles` collection by `author_id`, and then returns the row count for each group. That is, it computes the number of articles written by each author:
```
daniel-chambers, line 38:
```suggestion
This example partitions the `articles` collection by `author_id`, and then returns the row count for each group. That is, it computes the number of articles written by each author:
```

ndc-models/src/lib.rs
```diff
@@ -437,11 +461,130 @@ pub struct Query {
...
+#[schemars(title = "Group Order By Target")]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum GroupOrderByTarget {
+    Aggregate {
+        /// Aggregation method to apply
+        aggregate: Aggregate,
+        /// Non-empty collection of relationships to traverse
+        path: Vec<PathElement>,
+    },
+}
```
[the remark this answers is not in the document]
paf31: Ok this is added, please see what you think, thanks.

daniel-chambers (approved):
This is looking really good

PR #149, opened 2024-06-07
ndc-test: replay test folders in alphabetical order

hgiasac:
The `ndc-test replay` command doesn't execute snapshot folders in alphabet order. The current random testing behavior is only usable for mock requests and responses. It can't execute related mutations in sequence, e.g. `create -> update -> delete`.

### DIFF
--- a/ndc-test/src/lib.rs
+++ b/ndc-test/src/lib.rs
@@ -141,35 +141,45 @@ pub async fn test_snapshots_in_directory_with<
     snapshots_dir: PathBuf,
     f: impl Fn(Req) -> F,
 ) {
-    match std::fs::read_dir(snapshots_dir) {
+    match read_dir_sorted_by_name(snapshots_dir) {
         Ok(dir) => {
             for entry in dir {
-                let entry = entry.expect("Error reading snapshot directory entry");
+                if entry.metadata().is_ok_and(|md| md.is_dir()) {
+                    test!(
+                        entry.file_name().to_str().unwrap_or("{unknown}"),
+                        reporter,
+                        {
+                            async {
+                                let path = entry.path();
 
-                test!(
-                    entry.file_name().to_str().unwrap_or("{unknown}"),
-                    reporter,
-                    {
-                        async {
-                            let path = entry.path();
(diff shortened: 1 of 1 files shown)


ndc-test/src/lib.rs
```diff
@@ -141,37 +141,47 @@ pub async fn test_snapshots_in_directory_with<
...
         Err(e) => println!("Warning: a snapshot folder could not be found: {e}"),
     }
 }
 
+// Read directory and sort inner files by alphabet
+fn read_dir_sorted_by_name(snapshots_dir: PathBuf) -> std::io::Result<Vec<std::fs::DirEntry>> {
+    let mut paths: Vec<_> = std::fs::read_dir(snapshots_dir)?
+        .map(|r| r.expect("Error reading snapshot directory entry"))
+        .collect();
+    paths.sort_by_key(std::fs::DirEntry::path);
```
paf31, line 180:
```suggestion
    let mut paths: Vec<_> = std::fs::read_dir(snapshots_dir)?.collect()?;
    paths.sort_by_key(std::fs::DirEntry::path);
```

I think this ought to work?

hgiasac: No, it doesn't work. There are nested `Result` in `DirEntry` array. We need to unwrap twice

paf31: Oh, but then we'd have to change this line:
```
Warning: a snapshot folder could not be found: {e}
```

hgiasac: So what's your suggestion?
That error line is used to catch the `read_dir` function.

```rust
let mut paths: Vec<_> = std::fs::read_dir(snapshots_dir)?
```

paf31 (approved):
Thanks!

ndc-test/src/lib.rs
paf31: I just pushed a commit to use `exists()` for the warning, and to use `expect` once for everything else (which should be an error now)`.

Will merge as soon as CI finishes.

hgiasac: @paf31 feel free to update my PR. It's late in my place and I'm going to bed

PR #151, opened 2024-06-12
Remove duplication by setting values in the workspace file.

SamirTalwar:
We set the package version and edition in the workspace file.

We also use the workspace file to capture dependency versions, so that they are kept consistent across crates. The crates themselves still specify the features.


paf31 (approved):
Thanks!

PR #153, opened 2024-06-12
Remove RootCollectionColumn and ExistsInCollection::Unrelated

paf31:
As discussed in #146. If we agree on the approach, I will merge this branch there and we can tag a 0.2.0 release once engine is ready to work with multiple version ranges.

---

Notes:

- I removed `RootCollectionColumn`, planning to bring back the same functionality as a subset of "named scopes", proposed in #146. It's not currently used.
- I separated `ComparisonTarget` and `ComparisonValue` to make it possible to implement "filters by aggregates" (#147) without the undesirable side-effect of being able to put aggregates on both sides of a comparison operator.
- `ExistsInCollection::Unrelated` is unused, but also now useless without `RootCollectionColumn`, so it can be added back (if needed) alongside named scopes.

TODO:

- [ ] RFC
- [ ] Specification


daniel-chambers:
> ExistsInCollection::Unrelated is unused, but also now useless without RootCollectionColumn, so it can be added back (if needed) alongside named scopes.

I don't think this is true. I may want to define a permissions rule that says broadly, filter out rows if an AdminUsers table contains a row with the X-Hasura-UserId session variable value. That sort of filter doesn't require comparing anything in the root collection against the AdminUsers table.

A more compelling reason to remove it is that there's no feature in v3-engine that actually uses it right now, but we kinda know one is coming, right? So is it worth removing it, causing churn on connectors, only to add it back when we bring in named scopes?

Related comments about this from me here: https://github.com/hasura/ndc-spec/pull/146#issuecomment-2164250414

paf31:
I agree, I think we should keep `ExistsUnrelated` around. I'll update the PR.

paf31:
I will update this and move it to the other PR.

PR #154, opened 2024-06-15
ndc-test: add command to generate test fixtures 

hgiasac:
Add a command to generate test fixtures for queries and mutations that are used for `ndc-test replay`. This feature is ported from [ndc-sdk-go](https://github.com/hasura/ndc-sdk-go/tree/main/cmd/hasura-ndc-go#test-snapshots). I'm using it to quickly generate test cases for [Go-based connectors](https://github.com/hasura/ndc-stripe/tree/main/testdata/mock/mutation).

I also moved `common.rs` and `predicate.rs` modules outside because they are used for both query and mutation.

```sh
Usage: ndc-test generate [OPTIONS] --endpoint <ENDPOINT> --snapshots-dir <PATH>

Options:
      --endpoint <ENDPOINT>
          The NDC endpoint to test
      --seed <SEED>
          A 32-byte string with which to initialize the RNG
      --snapshots-dir <PATH>
          The directory used to store snapshot files
      --type [<STRING>...]
          The list of operation types to be generated. All operations will be generated by default [possible values: collection, function, procedure]
      --operation [<STRING>...]
          The list of query or mutation names to be generated. All operations will be generated by default
      --exclude-fields [<STRING>...]
          The list of field names to be excluded from the result type
      --exclude-arguments [<STRING>...]
          The list of argument names to be excluded from the input arguments
      --argument-depth <COUNT>
          The depth of nested arguments [default: 4]
      --field-depth <COUNT>
          The depth of nested selection fields in result type [default: 4]
      --write-mode <STRING>
          The write mode to be applied for existing files [default: ignore] [possible values: ignore, overwrite]
      --dry-run
          Generate random responses locally without fetching results from the connector endpoint
```


ndc-test/src/test_cases/fixture.rs
```diff
@@ -0,0 +1,467 @@
...
+            rng,
+        )?;
+        Ok(pred.map(|p| p.expr))
+    } else {
+        Ok(None)
+    }
+}
+
+#[allow(deprecated)]
+fn make_scalar_value(
```
paf31, line 285: Here is the problem with this approach for me, and why I didn't want to solve this here: there's no guarantee that we're generating semantically-meaningful values here. Just because a representation is string doesn't mean every string is meaningful, and just because a type is an array of strings doesn't mean every array of meaningful strings is meaningful.

I was planning to deal with this in a different way, by offering a way to generate test cases with missing values, but I'm not convinced it's worth it (compared to just hand-writing cases) given the amount of overhead for the user. 

For now I think this would be better as a separate tool or as part of the Go SDK's CLI, but if we continue it here, I'd like to hash out the approach in an RFC first.

hgiasac: That makes sense. I'm also wondering if we should generate random values of just static ones. 
However, the purpose of this command is for fixture generation. The connector author also needs to modify expected snapshot values. It's just convenient for nested field generation. Connector authors take hours or days to copy-paste hundreds of snapshots.   

It will be great if you have some ideas to improve this generator.

hgiasac:
I'm thinking if we define fixture presets for test case generation. Ideally, we can add an example field in schema type definition ([OAS 3 reference](https://swagger.io/docs/specification/adding-examples/)) fox documentation, or define the following JSON file:

```json
{
  "object_types": {
    "Foo": [
     {
        "foo": "bar"
      }
    ]
  },
  "scalar_types": {
    "Int": [1, 2, 3, 4],
    "Float": [1.0, 2.0, 3.0, 4.0],
  },
  "representations": {
    "boolean": [true],
    "biginteger": ["1000000", "2000000"],
    "string": ["foo", "bar"]
  }
}
```

The generator loads the preset file and picks random values in the order `object => scalar => representation`. 
@paf31 What do you think about this idea? Each connector can define a custom preset or users can also define it on their own.

paf31:
But I don't think it makes sense to say "42 is always a valid `Int`". Maybe for an output or for some input to a calculation, but maybe not as a primary key. Also, every combination of valid inputs is not always a valid invocation of a function.

This is why I use the data from the database to generate test cases, because there's no way they can be invalid, and we can even say in some cases when we expect queries to return non-empty result sets, etc.

paf31:
I'm going through and closing old PRs as we get ready for 0.2.0. I will leave the branch for reference but I don't think this is in line with the planned approach for ndc-test, and I maintain it's better handled in individual connectors or their CLI plugins. Next time, I'd suggest opening an RFC first to discuss an idea so that there's no risk of wasted work, but thanks for the PR anyway.

PR #155, opened 2024-06-17
Feature: named scopes

paf31:
[Rendered RFC](https://github.com/hasura/ndc-spec/blob/feature/scopes/rfcs/0015-named-scopes.md)

---

- [x] RFC
- [x] Specification updates
  - [ ] Changelog
  - [x] Specification pages
  - [x] Tutorial pages
  - [x] `reference/types.md`


rfcs/0015-named-scopes.md
```diff
@@ -0,0 +1,52 @@
...
+        name: String,
+        path: Vec<PathElement>,
+        ...
+        scope: Option<usize>,
+    },
+    ...
+}
+```
+
+Here, `scope` is an optional integer pointing to a position on a stack of scopes. As a connector enters a `Field::Relationship`, it should reset the stack to empty. As it enters an `Expression::Exists`, it should push the current row onto the stack.
```
daniel-chambers, line 44: Just for my clarity, what this is saying is that the scopes only exist with the confines of the Query predicate expression that they are instantiated in by using `Expression::Exists`, correct?
So, another predicate defined in a separate `Query` across a `Field::Relationship` can't see any scopes created in its parent-related `Query`'s predicate.

rfcs/0015-named-scopes.md
```diff
@@ -0,0 +1,52 @@
...
+
+So scope number zero refers to the current row in the current collection; scope number one refers to the row which was in scope in the immediately enclosing `EXISTS` expression; scope number two refers to the the row in scope in the next enclosing `EXISTS` expression, and so on. A query is only well-formed if these indices never meet or exceed the maximum number of enclosing `EXISTS` expressions up to the nearest enclosing `Query`.
```
daniel-chambers, line 46: Another way of expressing this is that the scope numbers are indexing into the scope ancestor chain from the current position, which is index 0. Index 1 is the current scope's parent, Index 2 is the current scope's grandparent, etc.

rfcs/0015-named-scopes.md
paf31: Yes, that's right.

ndc-models/src/lib.rs
```diff
@@ -649,6 +665,8 @@ pub enum ComparisonValue {
         /// Any relationships to traverse to reach this column
         #[serde(default)]
         path: Vec<PathElement>,
+
+        scope: Option<usize>,
```
daniel-chambers, line 669: Worth adding a `///` doc to?

daniel-chambers (approved):

ndc-models/src/lib.rs
paf31: Yes, thanks.

codedmart (approved):

PR #158, opened 2024-07-01
RFC: Remove relationship type

paf31:
- [x] [RFC](https://github.com/hasura/ndc-spec/blob/50fa28702b080429cdc8592f9a58e2964c1f53cd/rfcs/0018-remove-relationship-type.md)
- [ ] Specification updates
  - [ ] Changelog
  - [ ] Specification pages
  - [ ] Tutorial pages
  - [x] `reference/types.md`

### DIFF
--- a/ndc-models/src/lib.rs
+++ b/ndc-models/src/lib.rs
@@ -992,7 +992,6 @@ pub enum MutationOperation {
 pub struct Relationship {
     /// A mapping between columns on the source collection to columns on the target collection
     pub column_mapping: BTreeMap<FieldName, FieldName>,
-    pub relationship_type: RelationshipType,
     /// The name of a collection
     pub target_collection: CollectionName,
     /// Values to be provided to any collection arguments
--- a/ndc-models/tests/json_schema/mutation_request.jsonschema
+++ b/ndc-models/tests/json_schema/mutation_request.jsonschema
@@ -1322,7 +1322,6 @@
       "required": [
         "arguments",
         "column_mapping",
-        "relationship_type",
         "target_collection"
       ],
       "properties": {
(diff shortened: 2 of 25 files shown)


daniel-chambers:
Seems reasonable to me, but maybe some DB connector authors should chime in too. @soupi @hallettj?

daniel-chambers:
@paf31 I'm not sure I agree with the idea of sending incorrect information to the connector about the type of relationship. Maybe no connector is looking at that, but if we change from sending correct information to incorrect information, and they're relying on that field to change how they generate their query, it could have a material impact. And this is just this small field, what happens in the future with a more traumatic change than this?
 
Honestly, I'm starting to feel like we need to excise all ndc types from the engine IR and duplicate the IR->request mapping code for each NDC version. The engine IR would then be a superset of all ndc model versions and could map cleanly and correctly into both. Yes, there would be a lot of duplicate/similar code, but we'd avoid sending incorrect data in a potentially breaking way, and could provide 100% support for all versions of ndc (to their individual capabilities).

paf31:
> I'm starting to feel like we need to excise all ndc types from the engine IR and duplicate the IR->request mapping code for each NDC version

As a general strategy that makes sense to me.

I do think in this case, we could send `array` always, for the simple reason that the behavior is unspecified anyway, _but_ if the connector author decided that `object` meant "return the only row", we would be able to adjust the response to match that expectation on the engine side anyway. This reading is reasonable for anyone having used GDC before. If they decided it meant something else, I think it'd be reasonable to just say that's an incorrect assumption.

Also, in this case, we can just exhaustively look at the connectors that exist. That's not something we'll be able to do forever, of course, which is why I agree with your general point about IR generation being separate per version range. It's just a shortcut we have available this time if we want.

hallettj (approved):
The mongodb connector doesn't reference relationship type, so this seems like a fine change to me.

paf31:
There are parts of the spec which talk about paths consisting only of object relationships. Without this field, the connector would need to decide what that means, probably by reference to uniqueness constraints or something else. It seems cleaner to just have the caller indicate when uniqueness is expected.

PR #159, opened 2024-07-03
Borrowing for newtype wrappers

daniel-chambers:
This is just a quick fix to add [Borrow trait](https://doc.rust-lang.org/std/borrow/trait.Borrow.html) implementations for the new newtype wrappers so that the following works:

```rust
let arguments: &BTreeMap<ndc_models::ArgumentName, serde_json::Value> = todo!();
arguments.get("name") // can now pass a &str into a map and it works
```

Otherwise you have to write a special allocation:

```rust
arguments.get(&ndc_models::ArgumentName::from("movie_id"))
```

Also, the PR adds From instances to and from String for the double-wrapped newtypes, such as `FunctionName`.

### DIFF
--- a/ndc-models/src/lib.rs
+++ b/ndc-models/src/lib.rs
@@ -933,6 +933,42 @@ newtype! {ScalarTypeName over TypeName}
 newtype! {TypeName}
 newtype! {VariableName}
 
+impl From<String> for FunctionName {
+    fn from(value: String) -> Self {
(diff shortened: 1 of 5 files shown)


SamirTalwar (approved):

SamirTalwar:
Looks great. Perhaps we should add `ref_cast::RefCast` too.

paf31 (approved):

PR #160, opened 2024-07-03
RFC: Nested queries

paf31:
- [x] [RFC](https://github.com/hasura/ndc-spec/blob/feature/nested-queries/rfcs/0019-nested-collections.md)
- [x] Specification updates
  - [x] Changelog
  - [x] Specification pages
  - [ ] Tutorial pages
  - [x] `reference/types.md`


ndc-models/src/lib.rs
```diff
@@ -704,6 +704,16 @@ pub enum Field {
         #[serde(skip_serializing_if = "BTreeMap::is_empty", default)]
         arguments: BTreeMap<ArgumentName, Argument>,
     },
+    NestedCollection {
+        /// The name of the column which contains the nested data
+        column: FieldName,
+        /// Column arguments
+        arguments: BTreeMap<ArgumentName, Argument>,
+        /// Path to a nested array of objects contained within the selected column
+        field_path: Option<Vec<FieldName>>,
```
daniel-chambers, line 713: Do we need `field_path`, or could we use `Field::NestedField` to allow drilling through object paths? So instead of 

```yaml
CustomerAddresses:
  type: NestedCollection
  column: Customer
  arguments: {}
  field_path: ["Addresses"]
  query:
    aggregates:
      countryCount:
        type: ColumnCount
        column: Country
        distinct: true
```

we did
```yaml
Customer:
  type: Column
  column: Customer
  fields:
    type: Object
    fields:
      Addresses:
        type: NestedCollection
        column: Addresses
        arguments: {}
        query:
          aggregates:
            countryCount:
              type: ColumnCount
              column: Country
              distinct: true
```

Feels like the response shape more closely matches the nested object structure when it's done like this.

The other places we use `field_path` we don't actually care about the nesting because we're just using the result of traversing the nesting in a computation (eg predicates, ordering). But here, we do.

paf31: Good point, thanks, I don't think we need `field_path`.

specification/src/specification/queries/field-selection.md
```diff
@@ -25,8 +25,25 @@ If `fields` is omitted, the entire structure of the column's data should be retu
...
+
+For an array-typed column (whether nullable or not), the `fields` property may contain a `NestedField` with type `array`. 
+
+The `fields` property of the `NestedField` should contain _another_ `NestedField` structure, compatible with the type of the elements of the array. The selection function denoted by this nested `NestedField` structure should be applied to each element of each array.
+
+### Nested collections
+
+For a column whose type is an array of objects (whether nullable or not), the `fields` property may contain a `NestedField` with type `array_of_objects`.
+
+A connector should handle such fields by treating the nested array of objects as a collection. Such a field will include a nested `Query`, and the connector should execute that query in the context of this nested collection.
```
daniel-chambers, line 44: Probably worth saying something about needing to return a `RowSet` here.

ndc-models/src/lib.rs
```diff
@@ -679,13 +682,23 @@ pub struct NestedArray {
...
+// ANCHOR_END: ArrayOfObjects
+
 // ANCHOR: NestedField
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
 #[serde(tag = "type", rename_all = "snake_case")]
 #[schemars(title = "NestedField")]
 pub enum NestedField {
     Object(NestedObject),
     Array(NestedArray),
+    ArrayOfObjects(ArrayOfObjects),
```
daniel-chambers, line 701: We refer to this feature as "nested collections" everywhere except on the actual types... perhaps we should just change this to `NestedCollection(NestedCollection)`?

dmoverton: I agree that `NestedCollection` is a better name

daniel-chambers (approved):

PR #163, opened 2024-07-11
RFC: Add explicit enum member names for codegen purposes

GavinRay97:
In languages that lack support for declaring anonymous union types, like Java, it's not possible to codegen semantic names for `oneOf` members unless a name is a provided.

For example, given the following Rust definition:
```rs
enum Thing {
  Foo { name: str }
  Bar { id: int }
}
```

The Java representation can best be modeled as:
```java
sealed interface Thing {
  record Foo(String name) implements Thing {}
  record Bar(int id) implements Thing {}
}
```

To generate this from a JSON schema, we need to know what to call the enum members.
Currently, the generated schema doesn't provide this information:

https://github.com/hasura/ndc-spec/blob/78f52768bd02a8289194078a5abc2432c8e3a758/ndc-models/tests/json_schema/query_request.jsonschema#L116-L202

This change provides an additional `title` property, like:
```json
{
    "Aggregate": {
      "title": "Aggregate",
      "oneOf": [
        {
          "title": "ColumnCount",
          "type": "object",
          "required": [
            "column",
            "distinct",
            "type"
          ],
```

### DIFF
--- a/ndc-models/src/lib.rs
+++ b/ndc-models/src/lib.rs
@@ -241,21 +241,25 @@ pub struct ObjectField {
 #[schemars(title = "Type")]
 pub enum Type {
     /// A named type
+    #[schemars(title = "Named")]
     Named {
         /// The name can refer to a primitive type or a scalar type
         name: TypeName,
     },
     /// A nullable type
+    #[schemars(title = "Nullable")]
     Nullable {
         /// The type of the non-null inhabitants of this type
         underlying_type: Box<Type>,
     },
     /// An array type
+    #[schemars(title = "Array")]
     Array {
         /// The type of the elements of the array
         element_type: Box<Type>,
     },
     /// A predicate type for a given object type
+    #[schemars(title = "Predicate")]
     Predicate {
         /// The object type name
         object_type_name: ObjectTypeName,
@@ -269,8 +273,11 @@ pub enum Type {
 #[serde(tag = "type", rename_all = "snake_case")]
 #[schemars(title = "Comparison Operator Definition")]
 pub enum ComparisonOperatorDefinition {
+    #[schemars(title = "Equal")]
     Equal,
+    #[schemars(title = "In")]
     In,
+    #[schemars(title = "Custom")]
     Custom {
         /// The type of the argument to this operator
         argument_type: Type,
@@ -403,8 +410,10 @@ pub struct QueryRequest {
 #[schemars(title = "Argument")]
 pub enum Argument {
     /// The argument is provided by reference to a variable
+    #[schemars(title = "Variable")]
     Variable { name: VariableName },
     /// The argument is provided as a literal value
+    #[schemars(title = "Literal")]
     Literal { value: serde_json::Value },
 }
 // ANCHOR_END: Argument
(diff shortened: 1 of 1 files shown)


paf31:
Thanks for the detailed PR description :) I'm not opposed to this at all, but I'm curious - is this a peculiarity of whatever code generator you're using? Because technically Rust doesn't have anonymous sums either. The `type` field corresponds to the variant name.

paf31:
Oh also, you need to run `UPDATE_GOLDENFILES=1 cargo test`.

GavinRay97:
> but I'm curious - is this a peculiarity of whatever code generator you're using? Because technically Rust doesn't have anonymous sums either. The type field corresponds to the variant name.

I might be missing something, but in the `oneOf` object definitions, I can't seem to find any `type` name:
```json
{
"oneOf": [ 
     { 
       "type": "object", 
       "required": [ 
         "column", 
         "distinct", 
         "type" 
       ], 
       "properties": { 
         "type": { 
           "type": "string", 
           "enum": [ 
             "column_count" 
           ] 
         }, 
         "column": { 
           "description": "The column to apply the count aggregate function to", 
           "type": "string" 
         }, 
         "field_path": { 
           "description": "Path to a nested field within an object column", 
           "type": [ 
             "array", 
             "null" 
           ], 
           "items": { 
             "type": "string" 
           } 
         }, 
         "distinct": { 
           "description": "Whether or not only distinct items should be counted", 
           "type": "boolean" 
         } 
       } 
     }, 
     { 
       "type": "object", 
       "required": [ 
         "column", 
         "function", 
         "type" 
       ], 
       "properties": { 
         "type": { 
           "type": "string", 
           "enum": [ 
             "single_column" 
           ] 
         }, 
         "column": { 
           "description": "The column to apply the aggregation function to", 
           "type": "string" 
         }, 
         "field_path": { 
           "description": "Path to a nested field within an object column", 
           "type": [ 
             "array", 
             "null" 
           ], 
           "items": { 
             "type": "string" 
           } 
         }, 
         "function": { 
           "description": "Single column aggregate function name.", 
           "type": "string" 
         } 
       } 
     }, 
     { 
       "type": "object", 
       "required": [ 
         "type" 
       ], 
       "properties": { 
         "type": { 
           "type": "string", 
           "enum": [ 
             "star_count" 
           ] 
         } 
       } 
     } 
   ] 
}
```

The Rust definition has the names of these variants (`ColumnCount`, etc), but the generated schema does not:
```rs
pub enum Aggregate {
    ColumnCount {
        /// The column to apply the count aggregate function to
        column: FieldName,
        /// Whether or not only distinct items should be counted
        distinct: bool,
    },
    SingleColumn {
        /// The column to apply the aggregation function to
        column: FieldName,
        /// Single column aggregate function name.
        function: AggregateFunctionName,
    },
    #[schemars(title = "StarCount")]
    StarCount {},
}
```

GavinRay97:
Oh, I see what you mean now. `properties.type` is a single-value string enum that encodes the tagged-union type.

I don't see any reason why you couldn't use this in a codegen. IMO, it seems easier to read `title` than check whether there exists a `properties.type` and extract it -- but then there's the tradeoff between maintaining explicit `#[schemars(title = "ColumnCount")]` annotations which adds extra code debt and maintenance burden.

So I can see both pros and cons to this. Not opposed to closing it if other folks thing it's not worth the effort.

paf31:
@GavinRay97 Any more thoughts on this one? I'm inclined to close it since we technically have enough information in the current schema to accurately guide code generators.

PR #165, opened 2024-07-22
Exists in nested collections

paf31:
[Rendered RFC](https://github.com/hasura/ndc-spec/blob/feature/exists-in-nested-collection/rfcs/0020-exists-in-nested-collection.md)

---

- [x] RFC
- [x] Specification updates
  - [ ] Changelog
  - [x] Specification pages
  - [ ] Tutorial pages
  - [x] `reference/types.md`

### DIFF
--- a/ndc-reference/tests/query/predicate_with_exists_in_nested_collection/request.json
+++ b/ndc-reference/tests/query/predicate_with_exists_in_nested_collection/request.json
@@ -1,0 +1,63 @@
+{
+    "$schema": "../../../../ndc-models/tests/json_schema/query_request.jsonschema",
+    "collection": "institutions",
+    "arguments": {},
+    "query": {
+        "fields": {
+            "id": {
+                "type": "column",
+                "column": "id"
+            },
+            "name": {
+                "type": "column",
+                "column": "name"
+            },
+            "staff": {
+                "type": "column",
+                "column": "staff",
+                "arguments": {
+                    "limit": {
+                        "type": "literal",
+                        "value": null
+                    }
+                }
+            }
+        },
+        "predicate": {
+            "type": "exists",
+            "in_collection": {
+                "type": "nested_collection",
+                "arguments": {
+                    "limit": {
+                        "type": "literal",
+                        "value": null
+                    }
+                },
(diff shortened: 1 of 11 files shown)


ndc-reference/tests/query/predicate_with_exists_in_nested_collection/request.json
```diff
@@ -0,0 +1,63 @@
...
+                "type": "nested_collection",
+                "arguments": {
+                    "limit": {
+                        "type": "literal",
+                        "value": null
+                    }
+                },
+                "column_name": "staff"
+            },
+            "predicate": {
```
danieljharvey, line 38: Just checking my understanding - right now the example filters `last_name` containing "s", if we wanted to check for `specialities` that include an "s" instead, we'd nest another `exists` with a `nested_collection` here and put the `binary_comparison_operator` inside that?

ndc-reference/tests/query/predicate_with_exists_in_nested_collection/request.json
```diff
@@ -0,0 +1,63 @@
...
+                "operator": "like",
+                "value": {
+                    "type": "scalar",
+                    "value": "s"
+                }
+            }
+        }
+    },
+    "collection_relationships": {
+        "author_articles": {
```
danieljharvey, line 53: Am I missing something, or is this relationship definition not required for this request?

danieljharvey (approved):
A couple of clarifying questions for my understanding, this looks like it will do what we need though, great!

ndc-reference/tests/query/predicate_with_exists_in_nested_collection/request.json
paf31: Correct, but we can only use `exists` on nested _collections_, not nested arrays of strings. You would need to use something like a `contains` operator for your other example.

ndc-reference/tests/query/predicate_with_exists_in_nested_collection/request.json
paf31: Yes, thanks.

PR #169, opened 2024-08-08
Health endpoint is required not merely recommended

nullxone:
Minor change to indicate that the /health endpoint is required, not merely recommended (must not should) per [rfc 2119](https://datatracker.ietf.org/doc/html/rfc2119)

DX and control plane tooling relies on this endpoint to function.


paf31 (dismissed):
Thanks!

paf31 (approved):

PR #174, opened 2024-08-13
Format everything with Prettier, and check formatting in CI.

SamirTalwar:
Our JSON and Markdown files were inconsistently formatted. Prettier is pretty good at fixing this. I have run it over everything (ignoring a few things in `.prettierignore`) and added CI jobs to maintain formatting.

Most of the changes are whitespace:

  - 2 spaces, not 4, in JSON files
  - removing trailing spaces
  - adding a newline at the end of every file


paf31:
Honestly I've found this fairly frustrating working in engine this week, and I'm not sure I see the value here. Or at least it doesn't seem worth the effort to maintain.

My editor formatter rarely matches prettier, and the snapshot testing tool also doesn't seem to match in many cases, meaning I have to manually run the tool after updating tests in many cases.

What's the motivation for this, please?

paf31:
Manually rebasing this, it's easier than trying to resolve conflicts manually...

SamirTalwar:
@paf31: My motivation is not having to manually format Markdown files and instead just let the computer do it.

Most editors can be configured to run `prettier` on save (and it's the default in a few).

If you would prefer not to, then don't worry about it, but then I suggest reverting it on `main` to avoid merge conflicts.

PR #176, opened 2024-08-14
Update to http v1.

SamirTalwar:
`http` v1 is the stable release of the package. This has had many downstream effects, most notably on `axum` and `reqwest`.

This updates those two packages to pull in the new version of `http`.


hallettj (dismissed):
This looks good to me. I was looking into running the mongodb connector with these changes, but we'll have to update the connector for the 0.2 branch which we haven't done yet.

paf31 (approved):
Thanks!

PR #179, opened 2024-09-10
[NSP-12] New meanings for comparison and aggregate operators

paf31:
- [x] [RFC](https://github.com/hasura/ndc-spec/blob/a6610169f72cec6792d5e0830c57254e212b37d9/rfcs/0021-comparison-and-aggregate-meanings.md)
- [x] Specification updates
  - [x] Changelog
  - [x] Specification pages
  - [ ] Tutorial pages
  - [x] `reference/types.md`


ndc-models/src/lib.rs
```diff
@@ -330,10 +334,22 @@ pub enum ComparisonOperatorDefinition {
...
+        /// The scalar type of the result of this function, which should have
+        /// one of the type representations Int64 or Float64, depending on
+        /// whether this function is defined on a scalar type with an integer or
+        /// floating-point representation, respectively.
+        result_type: ScalarTypeName,
+    },
+    Custom {
+        /// The scalar or object type of the result of this function
+        result_type: Type,
+    },
```
daniel-chambers, line 352: No `Average` function? That was mentioned in the RFC.

ndc-models/src/lib.rs
```diff
@@ -330,10 +334,22 @@ pub enum ComparisonOperatorDefinition {
...
 #[schemars(title = "Aggregate Function Definition")]
-pub struct AggregateFunctionDefinition {
-    /// The scalar or object type of the result of this function
-    pub result_type: Type,
+pub enum AggregateFunctionDefinition {
+    Min,
+    Max,
+    Sum {
+        /// The scalar type of the result of this function, which should have
+        /// one of the type representations Int64 or Float64, depending on
```
daniel-chambers, line 344: This isn't always the case (Int64 or Float64). For example, the postgres connector's Float4 (ie 32-bit float) sum function returns Float4. However, all integer types do appear to sum to a 64-bit int.

paf31: Right, but a connector can always upcast a smaller int or float type to a 64-bit type, and since we can't control how many rows will be summed, we should pick the largest possible representations.

hallettj: I suppose I'll add that MongoDB has a Float128 type which I would want to use with inputs of that type

hallettj (approved):
Looks good to me!

ndc-models/src/lib.rs
0x777: > Right, but a connector can always upcast a smaller int or float type to a 64-bit type, and since we can't control how many rows will be summed, we should pick the largest possible representations.

Makes sense. [Here](https://substrait.io/extensions/functions_arithmetic/#sum) is how `sum is defined in substrait.

rfcs/0021-comparison-and-aggregate-meanings.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Future Work
+
+Other possible comparison operators (from https://datafusion.apache.org/user-guide/sql/operators.html#other-operators):
+
+- `IS DISTINCT FROM`, `IS NOT DISTINCT FROM`
+- `~`, `~*`, `!~`, `!~*` (regex-based matches, may be difficult to standardize across implementions, might consider "starts with", "contains" and "ends with" instead)
+- `@>`, `<@` (array operators)
+
+Other possible aggregate functions (from https://datafusion.apache.org/user-guide/sql/operators.html#comparison-operators):
```
0x777, line 75: Prior art on standardized aggregate functions: https://substrait.io/extensions/functions_arithmetic/#aggregate-functions.

0x777: If we add more aggregate functions to NDC, would that be a breaking change?

0x777 (dismissed):

ndc-models/src/lib.rs
daniel-chambers: This means that, in something like Postgres, we'd have to cast up any Float4 column to Float8 before aggregation in order to meet this spec. This could be seen as not respecting the choice of the end-user when they chose Float4 precision for their column. Postgres seems to deliberately retain that precision during aggregation (which it doesn't do for integers, only floats, from what I can see in our PG connector schema), probably because you can sum up pretty far using floats, you just lose precision, unlike integers where you overflow. 

Maybe overriding the chosen precision is okay, but it's something to consider.

rfcs/0021-comparison-and-aggregate-meanings.md
paf31: > If we add more aggregate functions to NDC, would that be a breaking change?

It would be breaking for engine since it would need to handle more cases. But if engine is checking the NDC version to make sure it's compatible with its own max-supported-version (not sure if we do this), then it's not a concern.

ndc-models/src/lib.rs
paf31: For 128-bit numeric types, unfortunately those wouldn't be supported by datafusion right now. We could still support custom aggregate operators for those types, but just not the standardized sum operator.

daniel-chambers:
OK, so I've updated this PR and it should be ready for final review and merge. cc @paf31 

* Added discussed average aggregate function to the spec
* Updated reference implementation with new comparison operators and aggregate functions
* Updated reference documentation with new aggregate function
* Updated tutorial documentation
* Updated changelog

Personally I still don't really like that we're forcing upcasting of lower precision types to higher precision types on aggregation, but I'm happy to start stricter like that and see if it causes issues for connectors and performance and we can relax it later if necessary.

ndc-reference/bin/reference/main.rs
```diff
@@ -1528,52 +1634,148 @@ fn eval_aggregate_function(
...
-                .try_into()
-                .map_err(|_| {
-                    (
-                        StatusCode::BAD_REQUEST,
-                        Json(models::ErrorResponse {
-                            message: "column value out of range".into(),
-                            details: serde_json::Value::Null,
-                        }),
-                    )
+    if let Some(first_value) = values.iter().next() {
```
paf31, line 1637: This function is quoted verbatim in the spec, so it's probably going to be a bit hard to follow now. You might want to break up the code a bit in the text.

daniel-chambers: OK, I've reduced the example in the docs down to just the first part of the function. The rest are basically the same just slight variants based on the type. The function isn't actually that big, it's just the shitty rust formatter that loves adding as many newlines as it possibly can blowing out every piece of error handling code into a lot of lines.

paf31 (commented):
Looks good, thanks!

codedmart (approved):

PR #183, opened 2024-10-03
[RFC] Nested Relationships

daniel-chambers:
~~**Note:** This PR is currently stacked on #185 and will be rebased once that PR has merged.~~ Rebased

[Rendered RFC](https://github.com/hasura/ndc-spec/blob/daniel/rfc-nested-relationships/rfcs/0024-nested-relationships.md)

- [x] RFC
- [x] Specification updates
  - [x] Changelog
  - [x] Specification pages
  - [x] Tutorial pages
  - [x] `reference/types.md`


hallettj (approved):
Looks good to me!

rfcs/0024-nested-relationships.md
```diff
@@ -0,0 +1,577 @@
+# Nested Relationships
+
+## Problem
+
+Imagine the following data model, in which we have three collections `Invoice`, `Suburb` and `Campaign`. The `Invoice` type has a nested object type `Addresses` that has foreign keys to the `Address` collection. `Invoice` also has a nested array of `Discount` objects, of which each has a foreign key to the `Campaign` collection. So `Invoice` can be said to have "nested relationships" to both `Address` and `Campaign`.
```
codingkarthik, line 5: Can this capability enable self-joining within a document?

For example:


```json
{
   "dispatched_orders": [
        {"order_id": 1},
        {"order_id": 2}
   ],
   "orders": [
         {"order_id": 1, "product_name": "ABC", "ordered_at": "2024-10-15"},

         {"order_id": 2, "product_name": "DEF", "ordered_at": "2024-10-15"}
     ]
}
```

Considering the above data, I want to retrieve all the dispatched orders along with the order details. 

P.S: BTW, I realize this is a terrible example because the data model could have a key called `dispatched` within every order item. I will think of a better example (if I can think one of) and add it to this thread.

rfcs/0024-nested-relationships.md
```diff
@@ -0,0 +1,577 @@
...
+
+  - `ComparisonValue::Column` - part of filter predicates; where the right hand side of a comparison operation references a column (this is currently unused in v3-engine, but is intended to be used in model permissions, as it was used for that purpose in v2 permissions)
+  - `OrderByTarget::Column` - when you want to order by a column across an object relationship ([example usage](#orderbytargetcolumn))
+
+  - `OrderByTarget::Aggregate` - when you want to order by an aggregate that happens across a nested object relationship ([example usage](#orderbytargetaggregate))
+
+  - `Dimension::Column` - when selecting a column to group by that occurs across a nested object relationship ([example usage](#dimensioncolumn))
+
+We are also unable to target a relationship at a collection where the target columns are nested. This is because `Relationship.column_mapping` does not allow for the target column to reference anything other than a column directly on the collection type. So if we wanted to follow the `Invoice.addresses.billingAddressId -> Address.addressId` object relationship in reverse, from Address (ie `Address.addressId -> Invoice.addresses.billingAddressId`), we couldn't.
```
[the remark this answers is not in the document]
daniel-chambers: That is exactly what this RFC proposes. 😄

rfcs/0024-nested-relationships.md
daniel-chambers: Not _really_, no. You can join back to the same collection of documents, so if your `orders.order_id` property was unique across all documents in the collection (ie no two documents contained the same `orders.order_id`), then yes it would work. A relationship always targets a collection; that is not changing with this RFC.

daniel-chambers: Actually, I take that back, it wouldn't work anyway, because `orders` is an array, and relationships can only target nested properties so long as they are nested objects, not nested arrays. We don't support implicit existential quantification over nested arrays at this time.

codingkarthik: Makes sense, thanks!

danieljharvey (dismissed):
Seems sensible, and we're keeping new functionality behind capabilities which should reduce the burden on connector implementors moving to `0.2.0`. 👍

codedmart (approved):

PR #185, opened 2024-10-15
Extends data in ndc-reference to capture a nested relationship

daniel-chambers:
This PR extends the test data in `ndc-reference` so that it contains a nested relationship. This is useful for #183, but is being done separately in this PR so as to isolate the golden test changes that happen as a result of the test data change, but not as a result of changes in that RFC.

`institutions` gains a `country_id` under `location`, and a `born_country_id` for each staff member. These new fields link to a new `countries` collection. A new institution has also been added.

Other changes:
* The `.json` data files have been renamed to `.jsonl` to capture the fact that they are JSON Lines files, not JSON files
* The `Int` scalar type has had `gt` and `lt` comparison operators added to it


danieljharvey (approved):

paf31 (approved):
Looks good.

As I mentioned on Slack, maybe let's take this opportunity to change the data for synthetic data too?

daniel-chambers:
If we do that, I'd like to do it after all the current PRs are merged (which are all stacked on top of this one), since all the current tests rely on this specific data, and I'd rather rework the tests and their results in one go after the changes are in to keep it separate.

PR #192, opened 2024-10-30
Use insta for reference connector tests

daniel-chambers:
This PR changes the reference connector snapshot tests to use `insta` rather than `goldenfile` to capture the snapshots. `insta` has a much better experience when it comes to diffing and maintaining the snapshot files.

All existing `expected.json` files in the reference connector tests have been changed into `expected.snap` files which are basically the same but just contain the `insta` header section. No actual test results have changed.


danieljharvey:
Pushed a commit to make running the tests in Nix pass. I'm not really sure why we bother natively building this library with Nix at all, rather than just using it to provide tooling, but we can look into that in future.

paf31 (approved):

daniel-chambers:
@danieljharvey Thanks for fixing that; I didn't notice that build had failed. 😢 I think the Nix flake exports the reference agent and the test suite. I can see some theoretical use for that where other connectors/the engine could use the flake to pull in the test suite/reference agent for integration testing purposes. However, I don't know if we actually do that anywhere in reality.

PR #193, opened 2024-10-31
Improve documentation around capabilities and fix broken includes in the docs

daniel-chambers:
This PR improves the documentation around capabilities by:
* annotating the comments on the ndc-model types to indicate where things are only used if certain capabilities are enabled
* mentioning the requirement for capabilities in the docs in various places where the capability was not discussed

It also fixes some broken includes in the docs that got broken in #192 where `.json` files got converted into `.snap` files. To ensure this doesn't happen again, a new CI build step has been added that builds the docs with mdbook and checks the outputted logs for errors.


ndc-models/src/expression.rs
```diff
@@ -112,14 +123,14 @@ pub enum ComparisonValue {
...
-    Scalar {
-        value: serde_json::Value,
-    },
-    Variable {
-        name: VariableName,
-    },
+    /// A scalar value to compare against
+    Scalar { value: serde_json::Value },
+    /// A value to compare against that is to be drawn from the query's variables.
+    /// Only used if the 'query.variables' capability is supported.
```
danieljharvey, line 132: As all the types become increasingly complex, this is going to be really helpful for new connector authors to know what they can initially stub out or mark with `todo!` etc.

danieljharvey (approved):

ndc-models/src/expression.rs
daniel-chambers: My thoughts exactly. 😄 I'm already losing track and I know the spec well.

PR #196, opened 2024-11-25
Upgrade smol_str to 0.3

jberryman:
Also upgrade schemars to our compatible fork that contains just a version bump.

This was motivated by: https://linear.app/hasura/issue/ENG-1279/try-to-move-to-rkyv since rkyv only supports smol_str 0.2 or 0.3 it seems...

- [ ] RFC
- [ ] Specification updates
  - [ ] Changelog
  - [ ] Specification pages
  - [ ] Tutorial pages
  - [ ] `reference/types.md`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,11 @@ return_self_not_must_use = { level = "allow" }
 too_many_lines = { level = "allow" }
 wildcard_imports = { level = "allow" }
 
+
+[patch.crates-io]
+# So we can bump the smol_str dependency, since rkyv only supports smol_str>=0.2 :(
+schemars = { git = 'https://github.com/hasura/schemars.git', branch = "v0" }
+
 [workspace.dependencies]
 async-trait = "0.1"
 axum = "0.7"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -25,6 +25,11 @@ return_self_not_must_use = { level = "allow" }
 too_many_lines = { level = "allow" }
 wildcard_imports = { level = "allow" }
 
+
+[patch.crates-io]
+# So we can bump the smol_str dependency, since rkyv only supports smol_str>=0.2 :(
+schemars = { git = 'https://github.com/hasura/schemars.git', branch = "v0" }
+
```
daniel-chambers, line 32: Will this actually work? From what I read in the cargo docs, this patch is not applied to consumers of this library, as all patches need to be defined at the top level.


> Remember that [patch] is applicable transitively but can only be defined at the top level so we consumers of my-library have to repeat the [patch] section if necessary.
> [Source](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#working-with-an-unpublished-minor-version)

jberryman: Hm I thought I needed this to build the library in isolation, but it seems not. anyway I realized I should probably keep this as a draft until I have something we'd want to merge in v3-engine

daniel-chambers: Seems to me like these patches should be applied in the engine codebase because that's where the deps need bumping to be able to use rkyv, no?

PR #203, opened 2025-02-10
Display the pretty difference when a test fails

codingkarthik:
- [ ] RFC
- [ ] Specification updates
  - [ ] Changelog
  - [ ] Specification pages
  - [ ] Tutorial pages
  - [ ] `reference/types.md`

### DIFF
--- a/ndc-test/src/error.rs
+++ b/ndc-test/src/error.rs
@@ -6,8 +6,6 @@ pub type OtherError = Box<dyn std::error::Error + Send + Sync>;
 pub enum Error {
     #[error("error communicating with the connector: {0}")]
     CommunicationError(#[from] Box<super::client::Error>),
-    #[error("error communicating with the connector: {0} ")]
-    CommunicationErrorWithContent(Box<super::client::Error>),
     #[error("error generating test data: {0}")]
     StrategyError(#[from] rand::Error),
     #[error("error parsing semver range: {0}")]
--- a/ndc-test/src/error.rs
+++ b/ndc-test/src/error.rs
@@ -5,9 +5,9 @@ pub type OtherError = Box<dyn std::error::Error + Send + Sync>;
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
     #[error("error communicating with the connector: {0}")]
-    CommunicationError(#[from] super::client::Error),
+    CommunicationError(#[from] Box<super::client::Error>),
     #[error("error communicating with the connector: {0} ")]
-    CommunicationErrorWithContent(super::client::Error),
+    CommunicationErrorWithContent(Box<super::client::Error>),
     #[error("error generating test data: {0}")]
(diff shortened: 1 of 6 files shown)


ndc-test/src/lib.rs
```diff
@@ -160,9 +164,14 @@ pub async fn test_snapshots_in_directory_with<
...
-                            let response = f(request).await?;
+                            let response = f(request).await;
+
+                            match response {
+                                Ok(response) => snapshot_test(snapshot_path, &response),
+                                Err(e) => Err(e),
+                            }
 
-                            snapshot_test(snapshot_path, &response)
+                            // snapshot_test(snapshot_path, &response)
```
paf31, line 174: Did you mean to leave this in?

ndc-test/src/snapshot.rs
```diff
@@ -61,17 +61,21 @@ impl<'a, C: Connector> Connector for SnapshottingConnector<'a, C> {
...
+        let compare: pretty_assertions::Comparison<R, R> =
+            pretty_assertions::Comparison::new(&snapshot, expected);
 
         if snapshot != *expected {
             let actual = serde_json::to_string_pretty(&expected)?;
+
             return Err(Error::ResponseDidNotMatchSnapshot(
                 snapshot_path.to_path_buf(),
                 actual,
+                format!("{compare}"),
```
paf31, line 78: Is there any use in storing this here, or can we just move the `format!` into the `Display` implementation for the error type?

ndc-test/src/error.rs
```diff
@@ -5,7 +5,9 @@ pub type OtherError = Box<dyn std::error::Error + Send + Sync>;
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
     #[error("error communicating with the connector: {0}")]
-    CommunicationError(#[from] super::client::Error),
+    CommunicationError(#[from] Box<super::client::Error>),
+    #[error("error communicating with the connector: {0} ")]
+    CommunicationErrorWithContent(Box<super::client::Error>),
```
paf31, line 10: Why do we have both of these?

ndc-test/src/lib.rs
codingkarthik: No, it's a stray comment now. I will remove it.

ndc-test/src/snapshot.rs
codingkarthik: So I did want to move into the error, but then compare's type is `pretty_assertions::Comparison<R, R>` and it would require extending parameterizing the `Error` type to `Error<R>`, which would be a substantial change, so for simplicity, I am converting it to String.

ndc-test/src/error.rs
codingkarthik: No reason to have both of these, will remove the latter one.

danieljharvey (approved):

PR #206, opened 2025-03-26
Query Relational types and capabilities

daniel-chambers:
- [ ] RFC
- [ ] Specification updates
  - [ ] Changelog
  - [ ] Specification pages
  - [ ] Tutorial pages
  - [ ] `reference/types.md`


ndc-models/src/relational_query/expression.rs
```diff
@@ -0,0 +1,911 @@
...
+    // to_timestamp_micros
+    // to_timestamp_millis
+    // to_timestamp_nanos
+    // to_timestamp_seconds
+    // today
+    // translate
+    // uuid
+
+    // Aggregate functions
+    Average {
```
paf31, line 728: We have both `Average` and `Mean`, we can remove one or the other.

ndc-models/src/relational_query/expression.rs
```diff
@@ -0,0 +1,911 @@
...
+    /// Only used when in specific contexts where the appropriate capability is supported:
+    /// * During projection: `relational_query.project.expression.scalar.date_part`
+    /// * During filtering: `relational_query.filter.scalar.date_part`
+    /// * During sorting:`relational_query.sort.expression.scalar.date_part`
+    /// * During joining: `relational_query.join.expression.scalar.date_part`
+    /// * During aggregation: `relational_query.aggregate.expression.scalar.date_part`
+    /// * During windowing: `relational_query.window.expression.scalar.date_part`
+    DatePart {
+        expr: Box<RelationalExpression>,
+        part: Box<RelationalExpression>,
```
paf31, line 330: This can only be a static string, and one of the same strings as used in `extract` in the group by part of the spec.

ndc-models/src/relational_query/expression.rs
daniel-chambers: Good catch, I have removed `Mean.

ndc-models/src/relational_query/expression.rs
daniel-chambers: I've added an enum and related capabilities for this.

ndc-models/src/capabilities.rs
```diff
@@ -25,9 +27,12 @@ pub struct LeafCapability {}
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
 #[schemars(title = "Capabilities")]
 pub struct Capabilities {
-    pub query: QueryCapabilities,
-    pub mutation: MutationCapabilities,
+    pub query: Option<QueryCapabilities>,
```
danieljharvey, line 30: Does this make this another breaking version change? Is there any way we could avoid it, at least in the short term whilst we're still very much experimenting?

I suppose the thing we want to avoid is engine sending a non-relational query to a connector that only does relational queries, but could we express that inside `RelationalQueryCapabilities` for the time being, and then make these top-level items optional once we're committed to this course of action?

daniel-chambers: Yeah, this is a good point, it would break things for old engines if they talk a connector that doesn't have query/mutation. Probably should revert this change and just leave it for when we stabilise this properly and then cut an NDC 0.3.0 version.

ndc-models/src/relational_query/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Hash, Serialize, Deserialize, JsonSchema)]
+#[serde(tag = "type", rename_all = "snake_case")]
+#[schemars(title = "Relation")]
+pub enum Relation {
+    From {
+        collection: CollectionName,
+        columns: Vec<FieldName>,
+    },
+    Paginate {
+        #[cfg(not(feature = "arc-relation"))]
```
danieljharvey, line 38: I like that this `Arc` / `Box` feature stuff is extremely explicit and inline.

ndc-models/src/relational_query/types.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Time stored as a signed 32bit int as seconds since midnight
+    Time32Second,
+    /// Time stored as a signed 32bit int as milliseconds since midnight
+    Time32Millisecond,
+    /// Time stored as a signed 64bit int as microseconds since midnight
+    Time64Microsecond,
+    /// Time stored as a signed 64bit int as nanoseconds since midnight
+    Time64Nanosecond,
+    /// Timestamp Second
+    TimestampSecond,
```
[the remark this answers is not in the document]
danieljharvey: Yeah, this makes sense, have made this change.

danieljharvey (approved):

danieljharvey:
This all looks fine to me, anything we should address more merging @paf31 ?

ndc-models/Cargo.toml
```diff
@@ -4,6 +4,10 @@ description = "Protocol definitions for the Hasura NDC specification"
 version.workspace = true
 edition.workspace = true
 
+[features]
+# Use Arc to nest Relations rather than Box
+arc-relation = []
```
paf31, line 9: Is there a reason we can't pick one or the other?

ndc-models/src/relational_query/expression.rs
```diff
@@ -0,0 +1,921 @@
...
+    /// * During projection: `relational_query.project.expression.aggregate.bool_or`
+    /// * During filtering: `relational_query.filter.aggregate.bool_or`
+    /// * During sorting:`relational_query.sort.expression.aggregate.bool_or`
+    /// * During joining: `relational_query.join.expression.aggregate.bool_or`
+    /// * During aggregation: `relational_query.aggregate.expression.aggregate.bool_or`
+    /// * During windowing: `relational_query.window.expression.aggregate.bool_or`
+    BoolOr {
+        expr: Box<RelationalExpression>,
+    },
+    Count {
```
paf31, line 749: I added `distinct: bool` here in response to a request from @tirumaraiselvan. I didn't add it for all of the other aggregate functions, but we could (although it's rarer in my experience). It would probably need another sub-capability though.

ndc-models/src/relational_query/notes.md
```diff
@@ -0,0 +1,14 @@
+# Changes
```
paf31, line 1: Did you mean to include this file? It looks a bit out of date but maybe it could become the RFC.

paf31 (changes requested):
Looks great but I noted a couple of small things. I can wrap this up if you like, just let me know. Thanks.

ndc-models/Cargo.toml
daniel-chambers: No connector will want to use `Arc`s for this. `Box` is appropriate. However, due to how these types are used in the engine, it requires `Arc` to avoid constant cloning due to the way DataFusion works. So this is a hack to allow engine to get Arcs but everybody else gets what this should be: a `Box`.

The alternative which I explored was have engine use `Arc<Mutex<..>>` so that it could avoid cloning (by using controlled mutation to move rather than clone), but that made the engine code gross. This seems like the least worst option that didn't allow weird engine concerns to pollute connectors.

ndc-models/src/relational_query/notes.md
daniel-chambers: No, sorry, this should be deleted, this was just my personal notes.

daniel-chambers:
> Looks great but I noted a couple of small things. I can wrap this up if you like, just let me know. Thanks.

Feel free to; I've handed this over to @danieljharvey to finish off as I'm needed elsewhere atm.

ndc-models/src/relational_query/expression.rs
danieljharvey: Makes sense, `distinct` added to match previous changes and `count_distinct` aggregation capability added.

ndc-models/src/relational_query/types.rs
```diff
@@ -0,0 +1,464 @@
...
+    /// Time stored as a signed 32bit int as milliseconds since midnight
+    Time32Millisecond,
+    /// Time stored as a signed 64bit int as microseconds since midnight
+    Time64Microsecond,
+    /// Time stored as a signed 64bit int as nanoseconds since midnight
+    Time64Nanosecond,
+    /// ISO 8601 timestamp
+    Timestamp,
+    /// Duration in seconds
+    DurationSecond,
```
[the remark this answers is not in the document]
danieljharvey: Makes sense, done.

PR #207, opened 2025-03-27
Add missing Millisecond enum in ExtractionFunctionDefinition type

hgiasac:
Add `Millisecond` enum in the `ExtractionFunctionDefinition` type

- [ ] RFC
- [ ] Specification updates
  - [ ] Changelog
  - [ ] Specification pages
  - [ ] Tutorial pages
  - [ ] `reference/types.md`

### DIFF
--- a/ndc-models/src/schema.rs
+++ b/ndc-models/src/schema.rs
@@ -230,6 +230,10 @@ pub enum ExtractionFunctionDefinition {
         /// The result type, which must be a defined scalar type in the schema response.
         result_type: ScalarTypeName,
     },
+    Millisecond {
+        /// The result type, which must be a defined scalar type in the schema response.
+        result_type: ScalarTypeName,
+    },
     Second {
         /// The result type, which must be a defined scalar type in the schema response.
         result_type: ScalarTypeName,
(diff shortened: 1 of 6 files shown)


daniel-chambers:
@hgiasac Not that I can't see the utility here, but what's the driving requirement for this? I have no context as to why this is important to you.

hgiasac:
Oh. So it is expected, right? I thought it was omitted by incident

hgiasac:
Just be curious why there are extraction functions for `nanoseconds` and `microseconds` but there is no function to extract milliseconds

daniel-chambers:
I think it was accidental.

Did you have a specific use case that called for it @hgiasac? We don't actually support using extraction functions in engine yet though, so the point might be moot for a while.

hgiasac:
@daniel-chambers I think I get OCD. I just feel that we need to fill that enum hole 😅 
However, I think that the extraction may be used for metrics such as extracting latency in milli or micro secs from timestamp

daniel-chambers (approved):

daniel-chambers:
@hgiasac Turns out this isn't unused in engine; PromptQL uses it. 😄

PR #209, opened 2025-04-22
Add `get_field` and `array_element` to relational expression

danieljharvey:
Allow accessing nested fields and array elements.

- [ ] RFC
- [ ] Specification updates
  - [ ] Changelog
  - [ ] Specification pages
  - [ ] Tutorial pages
  - [ ] `reference/types.md`
- [X] Does your feature add a new capability?
  - [X] Update `specification/capabilities.md` in the specification

### DIFF
--- a/ndc-models/src/relational_query/expression.rs
+++ b/ndc-models/src/relational_query/expression.rs
@@ -359,6 +359,18 @@ pub enum RelationalExpression {
         expr: Box<RelationalExpression>,
     },
     /// Only used when in specific contexts where the appropriate capability is supported:
+    /// * During projection: `relational_query.project.expression.scalar.get_field`
+    /// * During filtering: `relational_query.filter.scalar.get_field`
+    /// * During sorting:`relational_query.sort.expression.scalar.get_field`
+    /// * During joining: `relational_query.join.expression.scalar.get_field`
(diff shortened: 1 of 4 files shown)


paf31 (approved):

ndc-models/src/relational_query/expression.rs
```diff
@@ -358,6 +358,18 @@ pub enum RelationalExpression {
...
+    /// Only used when in specific contexts where the appropriate capability is supported:
+    /// * During projection: `relational_query.project.expression.scalar.get_field`
+    /// * During filtering: `relational_query.filter.scalar.get_field`
+    /// * During sorting:`relational_query.sort.expression.scalar.get_field`
+    /// * During joining: `relational_query.join.expression.scalar.get_field`
+    /// * During aggregation: `relational_query.aggregate.expression.scalar.get_field`
+    /// * During windowing: `relational_query.window.expression.scalar.get_field`
+    GetField {
+        column: Box<RelationalExpression>,
+        field: RelationalLiteral,
```
paf31, line 370: Would probably be worth introducing a separate type here if we're not going to allow arbitrary expressions, one which only allows `String` or `usize` for object and array indices respectively. No sense in indexing with a date, for example.

paf31 (approved):

PR #216, opened 2025-05-28
[PQL-416] Relational INSERT, UPDATE and DELETE IR

paf31:
I kept these as separate request and response types, instead of having some single "relational mutation request" or even just adding these to `MutationRequest` because it is experimental right now. Later we can consider bundling them into the existing `MutationRequest`.

- [ ] RFC
- [ ] Specification updates
  - [ ] Changelog
  - [ ] Specification pages
  - [ ] Tutorial pages
  - [ ] `reference/types.md`

### DIFF
--- a/ndc-models/src/relational_query/mod.rs
+++ b/ndc-models/src/relational_query/mod.rs
@@ -1,3 +1,5 @@
+use std::collections::BTreeMap;
+
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_with::skip_serializing_none;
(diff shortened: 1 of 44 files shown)


ndc-models/src/relational_query/mod.rs
```diff
@@ -25,6 +25,22 @@ pub struct RelationalQueryResponse {
     pub rows: Vec<Vec<serde_json::Value>>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+#[schemars(title = "RelationalInsert")]
+pub struct RelationalInsert {
+    pub collection: CollectionName,
```
paf31, line 32: TODO: do we want to deal with arguments?

ndc-models/src/requests.rs
```diff
@@ -243,3 +243,65 @@ pub struct ErrorResponse {
     pub details: serde_json::Value,
 }
 // ANCHOR_END: ErrorResponse
+
+// ANCHOR: RelationalInsertRequest
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "Relational Insert Request")]
+pub struct RelationalInsertRequest {
+    /// The name of the collection to insert into
+    pub collection: CollectionName,
```
paf31, line 252: TODO: should we handle arguments?

PR #227, opened 2025-06-26
Add `#[skip_serialize_none]` to MutationRequest and RelationalQuery

danieljharvey:
Make these match `QueryRequest` - we don't want to include empty request args.

- [ ] RFC
- [ ] Specification updates
  - [ ] Changelog
  - [ ] Specification pages
  - [ ] Tutorial pages
  - [ ] `reference/types.md`

### DIFF
--- a/rfcs/0024-nested-relationships.md
+++ b/rfcs/0024-nested-relationships.md
@@ -105,7 +105,6 @@ However, there are other contexts where we cannot navigate nested relationships.
 - `ExistsInCollection::Related` (used in filter predicates to navigate relationships) has no facility to descend into nested fields before navigating the relationship ([example usage](#existsincollectionrelated)).
 
 - `PathElement` is used to traverse relationships, but it does not have a nested field traversal facility. It is used in:
-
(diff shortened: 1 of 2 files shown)


ndc-models/src/relational_query/mod.rs
```diff
@@ -16,6 +16,7 @@ use crate::{ArgumentName, CollectionName, FieldName, OrderDirection};
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 #[schemars(title = "RelationalQuery")]
+#[skip_serializing_none]
```
BenoitRanque, line 19: This works, but I honestly prefer omitting the Option like we do elsewhere:

```rs
  // this goes on the property not the struct
  #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
```

codedmart (approved):

ndc-models/src/relational_query/mod.rs
danieljharvey: Yeah, that's what I'm used to, but this is how we do it everywhere else in this repo currently.

PR #243, opened 2025-08-19
Detect breaking changes

danieljharvey:
This adds a script that compares every `.jsonschema` file against the equivalent file in `main` using [json-schema-diff](https://docs.rs/json-schema-diff/latest/json_schema_diff/), and runs it in CI.

The package by its own admission is pretty limited, it found a breakage in a top-level `QueryRequest`, but not something deeper down in the chain. Still, this seems better than nothing, and at least makes our rules for additions more explicit.

I tried an NPM package too, but it exploded when it found a circular dep, which we have a lot of in Expression-shaped datatypes, and it [doesn't look like it's getting solved anytime soon](https://bitbucket.org/atlassian/json-schema-diff/issues/1/circular-references-should-not-result-in).

### DIFF
--- a/.github/workflows/breakage-checks.yml
+++ b/.github/workflows/breakage-checks.yml
@@ -1,0 +1,20 @@
+name: check for breaking changes 
+
+on:
+  push:
+
+jobs:
+  check-for-breaking-changes:
+    name: check for breaking changes
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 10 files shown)


danieljharvey:
Screenshot 2025-08-20 at 12 22 25

Still, better than nothing.

codedmart (approved):

